PR #5, opened 2022-07-08
Add repor formatting

mrnelson6:
**Sample**

**Summary**

**Tests**

**ArcGIS Maps SDK Version**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+Esri welcomes contributions from anyone and everyone. Please see our [guidelines for contributing](https://github.com/esri/contributing).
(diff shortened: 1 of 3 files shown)


arash-sh:
- This is fun 😄 
- Maybe add a UI with control guide? Could make the screen too crowded though. @mrnelson6  opinion?
- I personally feel that the controls for going up and down should be switched (I for up and k for down). But it may be just me
- When the restart button is clicked the sample selector UI is removed as well
- The text of the restart button is covering it so you can't press it when clicking over the text. Drag&drop the label onto the button to attach to it. Change the button color (not the background color) to white.
- We should look into why the map disappears when the plane is tilted

PR #9, opened 2022-07-18
Feature layer sample

mmgaw113:
**Sample**

Feature Layer Sample

**Summary**
Collects data through an HTTP request, parses through the data and creates objects for each data set. Positions the objects at the correct location in the world. Using a raycast, place the objects on the map so they are not floating. Users can select with object they would like to view in the selector on the left. 

**Tests**

Windows

**ArcGIS Maps SDK Version**

ArcGIS Maps SDK for Unreal Version 3572


arash-sh:
- Is there a reason for doing the query from game mode? It seems a bit odd that the query is made each time the game is started even if the feature layer is not loaded. 
One way to do it is to move the code to a C++ actor and place an instance of the actor in your scene. That way it only makes the query when the feature layer level gets loaded.  @mrnelson6 do you have any other suggestions?
If it has to be done from the game mode, three should be some way of limiting when the query is made.

- @mrnelson6 Please confirm, but I believe there shouldn't be any logs printed to the console except for warnings/errors. 
There are 2 UE_LOG statements in game mode

- In Feature Layer WBP, where the stadiums are spawned, an actor from the your level (FeatureLayer_lvl) should be assigned as the owner. If you end up making a new actor for the query, you can assign that actor. Otherwise, the map actor or the arcgis pawn can be used.

mrnelson6:
Ok yea that is all good feedback @arash-sh so I 2nd all that feedback. 

Additional stuff I would modify 
1. Put the stadium actors in a folder so that it stays more organized in Outliner. 
2. You are using the wrong basemap which doesn't require an API Key. Can you switch it to our default imagery one?
3.  The stadium dropdown is not in the correct location
image
4. I know @arash-sh has been looking into it for his sample, but whatever solution is discovered needs to be applied here to fix the issue where actors and UI are persisted when switching levels.
5. You have all these properties on the FeatureLayer blueprint for Lat/Lon/Altitude and other stuff. Instead of of duplicating latitude and longitude you should rely on the Location Component to hold that information.
6. Related to the previous point I would add blueprint properties for the team name, stadium name, and league so I can click on the actors and see that info like we are doing in Unity

mrnelson6:
@mmgaw113 not exactly sure what is going on here but if you just wanted the one commit from @arash-sh to make it so your UI disappeared you should have just cherry-picked that one commit. 
1. There is still a lot of routing stuff in here
2. This won't play nicely when inevitable you or Arash merge your branch

arash-sh:
This is the commit you need if you still want it:
"Remove the level-specific UIs when levels are loaded through the SampleSelector"
Or you could just wait until my branch is merged and pull main, but until then, the actors and UI will stay in the SampleViewer level

sample_project/Source/sample_project/sample_projectGameModeBase.cpp
```diff
@@ -15,4 +15,10 @@
 
 
 #include "sample_projectGameModeBase.h"
+#include "Json.h"
+void Asample_projectGameModeBase::StartPlay()
+{
+	Super::StartPlay();
+
+}
```
mrnelson6, line 23: Not necessary

sample_project/Source/sample_project/sample_projectGameModeBase.h
```diff
@@ -13,18 +13,24 @@
  * limitations under the License.
  */
 
+
```
mrnelson6, line 16: None of the changes in this file are necessary

sample_project/sample_project.uproject
```diff
@@ -7,7 +7,25 @@
...
+			"Enabled": true,
+			"SupportedTargetPlatforms": [
+				"Win64",
+				"Mac",
+				"Linux"
+			]
+		},
+		{
+			"Name": "ArcGISMapsSDK",
+			"Enabled": true
```
mrnelson6, line 28: Don't commit any changes to this file

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+// Called when the game starts or when spawned
+void AFeatureLayer::BeginPlay()
+{
+	Super::BeginPlay();
+	FHttpRequestRef Request = FHttpModule::Get().CreateRequest();
+	Request->OnProcessRequestComplete().BindUObject(this, &AFeatureLayer::OnResponseRecieved);
+	Request->SetURL("https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/Major_League_Baseball_Stadiums/FeatureServer/0/query?f=geojson&where=1=1&outfields=TEAM,NAME,LEAGUE");
```
mrnelson6, line 58: Store the featureserver url in a variable so it is apparent to users who want to repurpose this for another purpsose. Same thing with the outfields

mrnelson6 (changes requested):
Also adjust the layout on your blueprints

sample_project/sample_project.uproject
```diff
@@ -6,8 +6,7 @@
 	"Modules": [
 		{
 			"Name": "sample_project",
-			"Type": "Runtime",
-			"LoadingPhase": "Default"
```
mrnelson6, line 10: Why does this need to be deleted?

sample_project/Source/sample_project/sample_projectGameModeBase.h
```diff
@@ -19,12 +19,9 @@
 #include "GameFramework/GameModeBase.h"
 #include "sample_projectGameModeBase.generated.h"
 
-/**
- * 
- */
-UCLASS()
+
+UCLASS(Blueprintable)
```
mrnelson6, line 23: Why are these changes needed?

mmgaw113: Should UCLASS() be there? Or should it just not have Blueprintable?

mrnelson6: The blueprintable. ideally this file should not be showing up on your PR because there should be 0 changes to it.

mmgaw113: I copied and pasted what was in the file from main. If it still shows up with changes I will probably need help figuring it out.

mrnelson6 (changes requested):
One last thing, what changed with the ThirdPersonCharacter level in this PR?

mmgaw113 (commented):
I genuinely don't know how 3rd person map got changed or what the change was. I cloned the project again and moved the entire third person folder into my project and pushed and it doesn't show the issue anymore. If you have time, I would really like to in depth go through pull requests, and what you expect out of them, on Monday because I have never done them and want to get better at them so we don't go through this with my next sample.

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -0,0 +1,68 @@
...
+void AFeatureLayer::BeginPlay()
+{
+	UWebLink* url = NewObject<UWebLink>();
+	url->link = "https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/Major_League_Baseball_Stadiums/FeatureServer/0";
+	url->requestHeaders = "f=geojson&where=1=1";
+	url->outFieldHeader = "outfields=TEAM,NAME,LEAGUE";
+
+	url->requestHeaders += "&" + url->outFieldHeader;
+	url->link += "/Query?" + url->requestHeaders;
+	Super::BeginPlay();
```
mrnelson6, line 62: Move this to the top of the function

mrnelson6 (approved):
Looks good you are good to merge

PR #12, opened 2022-08-04
Geocoding sample

arash-sh:
**Sample**
Geocoding: Enter an address to get the geographic location of it. The pawn will be moved to that location and show a visual cue. 
Reverse geocoding: Shift+ left click on a location to see the address. The queries are made through the Geocoding service API.

**Summary**
Enter an address on the text field in the top and the camera will move to that location and show a pin on that point as well as the address returned by the API.  Ray casts will determine the elevation of the point.
Clicking on a location while holding the shift will make a query to get the address/description of that point. A visual cue shows the address. The scale of the visual cue is set based on the distance from the camera and the text will face the camera. 
No buildings are added to the map, since the buildings may occlude the text shown to the user.

**Tests**
Windows

**ArcGIS Maps SDK Version**
1.0.0


sample_project/Content/SampleViewer/Samples/Geocoding/readme.md
```diff
@@ -0,0 +1,16 @@
+Find the location for an address or find the address of a selected location using the Geocoding service. 
+
+![Geocoding](Geocoding.jpg)
+
+## How it works
+
+1. Set your API Key in the ArcGISMapActor using the World Outliner window (if you are using the SampleViewerLevel you can also set the key through the UI or the level blueprint).
+2. Enter an address in the text field to get the location of the address.
+2. While holding shift, left-click on a location on the map to get the address (or description) of that location.
```
Jade-JadeH, line 9: numbering

sample_project/Content/SampleViewer/Samples/Geocoding/readme.md
```diff
@@ -0,0 +1,16 @@
...
+3. Raycasts are used to determine the height at the queried location to account for elevation.
+4. This sample is only set up to work with mouse and keyboard.
```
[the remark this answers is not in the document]
arash-sh: I think in general we should have two separate sections. 1 for how to use the sample, and 1 for how to reproduce it/how it works. Similar to some other runtime samples (e.g., https://github.com/Esri/arcgis-runtime-samples-dotnet/tree/main/src/UWP/ArcGISRuntime.UWP.Viewer/Samples/Analysis/DistanceMeasurement)
But if we do that we should do it for all samples

sample_project/Source/sample_project/Geocoding/QueryLocation.h
```diff
@@ -0,0 +1,68 @@
...
+public:	
+	AQueryLocation();
+	
+	virtual void Tick(float DeltaTime) override;
+
+	void SetupAddressQuery(UArcGISPoint* InPoint, FString InAddress);
+	void SetupLocationQuery(FVector3d InPoint);
+	void UpdateAddressCue(FString inAddress);
+	
+	UPROPERTY(VisibleAnywhere, Category = "ArcGISMapsSDK|SampleDefaultPawn")
```
mrnelson6, line 42: You should remove the category from this

sample_project/Source/sample_project/Geocoding/Geocoder.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+// Parse the response for a geocoding query
+void AGeocoder::ProcessAddressQueryResponse(FHttpRequestPtr Request, FHttpResponsePtr Response, bool bConnectedSucessfully)
+{
+	FString ResponseAddress = "";
+	TSharedPtr<FJsonObject> JsonObj;
+	TSharedRef<TJsonReader<>> Reader = TJsonReaderFactory<>::Create(Response->GetContentAsString());
+	
+	// Check if the query was successful
+	if (FJsonSerializer::Deserialize(Reader, JsonObj) &&
```
mrnelson6, line 95: Just how you added logic in for the routing sample do you want to add a case to handle an error with either:
1. There is an invalid api key
2. The api key does not have geocoding permissions?

mrnelson6 (changes requested):

Jade-JadeH:
very nice. However, when there're multiple possible locations if user enters an incomplete address and hit enter, it takes user to one of them and you cannot select from dropdown to choose which one is the intended address. (see screenshots below). Can you make a dropdown with a sorted list of address? You don't need to have all addresses listed, you can return the first 5 from the sorted list. If this is too involved, perhaps you could prevent from showing results before matching user's entering with the address 

intended:
image

Actual result returned: 
If entering: 112 West Hamilton:
image
if entering: 112 West Hamilton, PA 
image

arash-sh:
> Can you make a dropdown with a sorted list of address? You don't need to have all addresses listed, you can return the first 5 from the sorted list. If this is too involved, perhaps you could prevent from showing results before matching user's entering with the address

I looked into this, and with the way the server responds, showing the top 5 would not be a good solution I think. For your example of "112 west Hamilton", the server returned 35 possible candidates, none of which is in State College. So if we give the top 5 (or even all 35), it would not help the user.
The service also returns a score of each candidate, which can be pretty high. In this example the top 3 candidates have a score of 98.21%. The top 10 have a score higher than 97%. So if we want to prevent showing the results until the intended location is found, we can't really rely on the score because the wrong candidates can still have a high score (even 100% doesn't guarantee correct answer, because there can be multiple candidates with 100% and it is possible that none of the is what the user intended).
I think as an alternative, we can show a message like in this image. The message will be shown when there are more than one candidates (but I think that will be very often the case even if the address is very specific). 
What do you think?
temp

sample_project/Source/sample_project/Geocoding/Geocoder.cpp
arash-sh: This will be included in the next push. The above errors are only relevant for the address search since the reverse geocoding doesn't require an API. 
The reverse geocoding would still show an error if the service sent an error message. The only error that I could force was when the lat/lon are invalid, but that will (hopefully) not happen in our use case

mrnelson6 (approved):
Nice PR @arash-sh ! Good to merge

PR #17, opened 2022-08-05
Routing sample

arash-sh:
**Sample**

Improved routing sample 

**Summary**
- UI shows error messages returned from the routing service. Only pertains to error messages that are contained in the response from server (i.e., token invalid, or token not configured for routing service). Errors related to communication, server, etc. are ignored.
- If the raycast hits something other than the map actor, it will be ignored (no marker spawned)  

**Tests**
Windows

**ArcGIS Maps SDK Version**
1.0.0


mrnelson6 (approved):
Nice improvements

Jade-JadeH:
image
1. I still don't see token invalid message?
2. What is the grey sphere on the map?

arash-sh:
> 2. What is the grey sphere on the map?

The sphere is the default pawn that gets created when the sample viewer is loaded. It is not related to the sample and has to be changed in the sample viewer (if at all)

Jade-JadeH (approved):

Jade-JadeH:
> The sphere is the default pawn that gets created when the sample viewer is loaded. It is not related to the sample and has to be changed in the sample viewer (if at all)

Is there a reason why it's visible? If not, can we just hide it?

PR #19, opened 2022-08-17
Flight simulator sample - Physics update

PauFlorenti:
**Sample**

Updating the Flight Simulator sample in its branch.

**Summary**
* Reimported ship mesh to fix orientation.
* Ship movement done through physics.
* Created a simple HUD showing where the ship is heading.
* Changed controls and added mouse control support for pitch and roll.
* Improved crash widget, allowing to quit the sample.
* Updated readme.md for Flight Simulator sample.

**ArcGIS Maps SDK Version**
1.0.0


mrnelson6:
Ok a couple things of feedback. Overall looks good.
1. There is an API Key in your level. Can you remove that?
2. Can you make it so the ship has a higher top speed and a faster acceleration? I think it would be cooler if the pilot was able to cover larger distances. With that being said I noticed something weird when you went a distance from the start where the airplane disappeared do you know what is up with that?
3. We want the sub levels to be linked to the parent SampleViewer level. So when you launch the sample viewer level it is visible on the side bar. To accomplish this you need to first click Window->Levels-> and Add the flight sim level. and the other thing you need to do is add another entry in the SamplesDataTable for your level

PauFlorenti:
API key removed and top speed increased to x3 what it previously was.
I could not reproduce the airplane disappearing. Does it happen straight from the track at a distance? Or by doing a specific path or movement?
Level has been added to the sampler viewer level, but I found a few issues when restarting the level after crashing the plane regarding the API Key.

PauFlorenti:
@arash-sh  I've found an issue working in this sample. When loading the level from the sample viewer level, the mouse is not being captured for the movement of the ship. Any leads on why the mouse is not being captured?

arash-sh:
> @arash-sh I've found an issue working in this sample. When loading the level from the sample viewer level, the mouse is not being captured for the movement of the ship. Any leads on why the mouse is not being captured? 

In the SampleViewer level, the mouse is only captured when a mouse button is pressed. I think the controls should be changed in so that it behaves the same way in the standalone level. I also noticed that the SampleSelector UI is removed when the flight sim level is loaded. Is that OK @mrnelson6 ? 

Other comments:
- Possibly cap the rotation of the spaceship so that it doesn't become upside-down
- Shift button is a bit difficult to hold while pressing WASD. Possibly use W and S for acceleration/deacceleration since the mouse is already setting the pitch of the air plane. Alternatively, remove the controls for mouse so that both hands are on the keyboard and the right shift key (or any other key) can be used for acceleration
- When the level is restarted after crash, the API key should be re-applied to the map (e.g., the same way the SampleSelector does it)
- A unit for the speed in the UI would be good
- I think the way it is currently implemented, the altitude value would not be correct in a global scene.

arash-sh:
I don't know if it makes sense from the physics/realism point of view, but for me it feels more natural if the keys 'e' and 'q' only change the yaw rather than the whole rotation. Right now, for example, if the plane is tilted upwards and you hold 'e' it will end up pointing downwards. You can change this in the UpdateYaw function by using the up vector of the ENU transformation without multiplying it with the rotation of the ship mesh. But you should get the opinion of others on this as well. 

In the UpdateYaw/Pitch/Roll functions, I don't think the Lerp has an effect since the alpha value is constant. You can remove the lerp and adjust the speed value instead.

I wasn't able to run the flight simulator level (without loading it through sample selector). The plane falls through the surface before the map is loaded. I think you should place the Ship_BP in the scene instead of using the game mode 

When the plane crashes and the level is restarted there is a blueprint runtime error in the SampleSelector

If I remember right, this sample should use a global scene. This may cause some new problems that need to be fixed.

PauFlorenti:
The ship exploding at the beginning of the level is due to the elevation being enabled in the MapComponent, so to avoid the ship crashing at the beginning I am delaying the elevation of the map for a few seconds. The ideal would be to have a way for runtimecore to tell us if the map is fully loaded and then spawn the ship and start flying. Until that is not added we will stick with this solution.
- Disabling collisions for a few seconds may be another solution.

sample_project/Content/SampleViewer/Samples/FlightSimulator/readme.md
```diff
@@ -15,7 +15,8 @@ Allows users to explore a city/map from the perspective of a Third Person Charac
 
 ## Controls
 This sample supports keyboard controls and use of an Xbox/Playstation controller
-Keyboard Controls: W for Acceleration, S for Deceleration, A for Yaw Left, D for Yaw Right, K or 5 (on numberpad) for pitch Up, I or 8 (on numberpad) for pitch down, J or 4 (on numberpad) for roll left, L or 6 (on numberpad) for roll right. H can be used to toggle the landing gear.
+Keyboard Controls: Left Shift for Acceleration, Q for Yaw Left, E for Yaw Right, W or Mouse Y for pitch Up, S or Mouse Y for pitch down, A or Mouse X for roll left, D or Mouse X for roll right.
```
airaolagoitia, line 18:
```suggestion
Keyboard Controls: Spacebar for Acceleration, Q for Yaw Left, E for Yaw Right, W or Mouse Y for Pitch Up, S or Mouse Y for Pitch Down, A or Mouse X for Roll Left, D or Mouse X for Roll Right. 
```

sample_project/Content/SampleViewer/Samples/FlightSimulator/readme.md
```diff
@@ -15,7 +15,8 @@ Allows users to explore a city/map from the perspective of a Third Person Charac
 
 ## Controls
 This sample supports keyboard controls and use of an Xbox/Playstation controller
-Keyboard Controls: W for Acceleration, S for Deceleration, A for Yaw Left, D for Yaw Right, K or 5 (on numberpad) for pitch Up, I or 8 (on numberpad) for pitch down, J or 4 (on numberpad) for roll left, L or 6 (on numberpad) for roll right. H can be used to toggle the landing gear.
+Keyboard Controls: Left Shift for Acceleration, Q for Yaw Left, E for Yaw Right, W or Mouse Y for pitch Up, S or Mouse Y for pitch down, A or Mouse X for roll left, D or Mouse X for roll right. 
+H can be used to toggle the landing gear.
```
airaolagoitia, line 19: This doesn't work for me. Or I don't see anything happening.

sample_project/Config/DefaultEngine.ini
```diff
@@ -4,6 +4,7 @@
 GameDefaultMap=/Game/SampleViewer/SampleViewerLevel.SampleViewerLevel
 EditorStartupMap=/Game/SampleViewer/SampleViewerLevel.SampleViewerLevel
 GlobalDefaultGameMode=/Script/sample_project.sample_projectGameModeBase
+GameInstanceClass=/Game/SampleViewer/BP_GameInstance.BP_GameInstance_C
```
airaolagoitia, line 7: will this affect other samples?

sample_project/Config/DefaultInput.ini
```diff
@@ -81,12 +81,12 @@ DoubleClickTime=0.200000
 +ActionMappings=(ActionName="Crouch",bShift=False,bCtrl=False,bAlt=False,bCmd=False,Key=C)
 +ActionMappings=(ActionName="Crouch",bShift=False,bCtrl=False,bAlt=False,bCmd=False,Key=Gamepad_FaceButton_Left)
 +ActionMappings=(ActionName="PlaceRoutePoint",bShift=True,bCtrl=False,bAlt=False,bCmd=False,Key=LeftMouseButton)
-+AxisMappings=(AxisName="Thrust",Scale=1.000000,Key=W)
-+AxisMappings=(AxisName="Thrust",Scale=-1.000000,Key=S)
++ActionMappings=(ActionName="Thrust",bShift=False,bCtrl=False,bAlt=False,bCmd=False,Key=SpaceBar)
```
airaolagoitia, line 84: same. can these affect other samples?

airaolagoitia (commented):
It would be nice to improve the "flight dynamics" like when you want to go right, you'll press the right key (D or Arrow Right) and the plane will yaw to the right while rolling to the right as well. When you stop pressing the right key, it would stop rolling.

Do you think that is easy to implement?

sample_project/Config/DefaultInput.ini
PauFlorenti: As far as I tested, this action mappings should not affect other samples since are only used in the flight simulator sample.

PauFlorenti:
> It would be nice to improve the "flight dynamics" like when you want to go right, you'll press the right key (D or Arrow Right) and the plane will yaw to the right while rolling to the right as well. When you stop pressing the right key, it would stop rolling.

Do you mean to roll and change pitch at the same time with just pressing a key? I don't think that is the way a plane should move, I believe we should give the option to roll, pitch and yaw independently.
We could add another key to do both roll and pitch at the same time though

airaolagoitia (approved):

airaolagoitia:
I think you should merge unless somebody else has comments. @mrnelson6 can review when he's back.

sample_project/Content/SampleViewer/Samples/FlightSimulator/readme.md
```diff
@@ -7,15 +7,20 @@ Allows users to explore a city/map from the perspective of a Third Person Charac
...
-6. Enjoy flying around the world.
+2. Make sure the component is in **Global Mode**.
+3. Pick an airport to take off from and set the latitude and longitude of the airport as your world center
+4. Create a new blueprint and call it "bp_Runway".
+5. Add a cube and change the scale of it to be 50, 2500, 1. Or something similar that suits your needs.
+6. Add your Plane blueprint on top of your run way.
+7. Enjoy flying around the world.
+
+Movement of the plane is done through Unreal physics.
+We add a physics **linear velocity** depending on the thrust and rotate for each of the axis adding a **torque** in such axis so that the plane can rotate in the roll, pitch and yaw axes. If the plane reduces enough the speed, a false gravity drags the plane towards Earth.
```
tTakeda09, line 18:
```suggestion
2. Set the **Origin Position** for the place where you want to place the runway.
3. Create a runway and add it to the level. In this sample, the runway is created by using a cube and saved as **BP_Runway**.
4. Create a flying object as a **Pawn** and attach the [**ArcGIS Camera Component**](https://developers.arcgis.com/unreal-engine/maps/camera/#arcgis-camera-component) and [**ArcGIS Location Component**](https://developers.arcgis.com/unreal-engine/maps/location-component/).
5. Connect the flying object to the [**Player Controller**](https://docs.unrealengine.com/5.0/en-US/player-controllers-in-unreal-engine/) and place it on the map. In this sample, the flying object is saved as **BP_Ship**.
6. Work on [physics](https://docs.unrealengine.com/5.0/en-US/physics-in-unreal-engine/) from Unreal Engine's feature for the flying object. This sample uses [**Set Physics Linear Velocity**](https://docs.unrealengine.com/5.0/en-US/BlueprintAPI/Physics/SetPhysicsLinearVelocity/) for the speed control and [**Add Torque in Degrees**](https://docs.unrealengine.com/4.26/en-US/BlueprintAPI/Physics/AddTorqueinDegrees/) for the object's roll, pitch, and yaw. If the plane reduces enough its speed, false gravity drags the plane toward Earth.
```

This is how it looks like to me though I might be wrong...

sample_project/Content/SampleViewer/Samples/FlightSimulator/readme.md
```diff
@@ -7,15 +7,20 @@ Allows users to explore a city/map from the perspective of a Third Person Charac
...
 
 ## Controls
 This sample supports keyboard controls and use of an Xbox/Playstation controller
-Keyboard Controls: W for Acceleration, S for Deceleration, A for Yaw Left, D for Yaw Right, K or 5 (on numberpad) for pitch Up, I or 8 (on numberpad) for pitch down, J or 4 (on numberpad) for roll left, L or 6 (on numberpad) for roll right. H can be used to toggle the landing gear.
+Keyboard Controls: Spacebar for Acceleration, Q for Yaw Left, E for Yaw Right, W or Mouse Y for Pitch Up, S or Mouse Y for Pitch Down, A or Mouse X for Roll Left, D or Mouse X for Roll Right. 
+H can be used to toggle the landing gear.
 Controller Controls: Right Trigger for Acceleration, Left Trigger for Deceleration, Left Stick Left and Right for Yaw, Right Stick Left and Right for Roll, Right Stick up and down for Pitch. Right Thumbstick button to toggle landing gear.
```
tTakeda09, line 24: I added the control to the How to use the sample section above.

tTakeda09 (changes requested):
I checked readme. It is just a suggestion and I might be wrong. So if there is any better way of explaining the sample, feel free to make your version.

sample_project/Content/SampleViewer/Samples/FlightSimulator/readme.md
```diff
@@ -1,21 +1,35 @@
...
+6. Work on [physics](https://docs.unrealengine.com/5.0/en-US/physics-in-unreal-engine/) from Unreal Engine's feature for the flying object. This sample uses [**Set Physics Linear Velocity**](https://docs.unrealengine.com/5.0/en-US/BlueprintAPI/Physics/SetPhysicsLinearVelocity/) for the speed control and [**Add Torque in Degrees**](https://docs.unrealengine.com/4.26/en-US/BlueprintAPI/Physics/AddTorqueinDegrees/) for the object's roll, pitch, and yaw. If the plane reduces enough its speed, false gravity drags the plane toward Earth.
 
 ## Controls
 This sample supports keyboard controls and use of an Xbox/Playstation controller
-Keyboard Controls: W for Acceleration, S for Deceleration, A for Yaw Left, D for Yaw Right, K or 5 (on numberpad) for pitch Up, I or 8 (on numberpad) for pitch down, J or 4 (on numberpad) for roll left, L or 6 (on numberpad) for roll right. H can be used to toggle the landing gear.
+Keyboard Controls: Spacebar for Acceleration, Q for Yaw Left, E for Yaw Right, W or Mouse Y for Pitch Up, S or Mouse Y for Pitch Down, A or Mouse X for Roll Left, D or Mouse X for Roll Right. 
+H can be used to toggle the landing gear.
```
PauFlorenti, line 32:
```suggestion
```

sample_project/Content/SampleViewer/Samples/FlightSimulator/readme.md
```diff
@@ -1,21 +1,35 @@
...
 Controller Controls: Right Trigger for Acceleration, Left Trigger for Deceleration, Left Stick Left and Right for Yaw, Right Stick Left and Right for Roll, Right Stick up and down for Pitch. Right Thumbstick button to toggle landing gear.
```
PauFlorenti, line 33:
```suggestion
```

tTakeda09 (approved):

airaolagoitia:
@mrnelson6 can you give this a look before we merge it?

Jade-JadeH:
I'm checking this PR today to see if we should close or keep it. but was not able to compile this with 1.0.0 (which is what this was built for). @PauFlorenti Do you think it's worthwhile to make this to work with the current version? 

image

airaolagoitia:
@Jade-JadeH that error is probably due to an unclean setup

Jade-JadeH:
@airaolagoitia I did try a clean setup by removing all unnecessary folders and the VS file

airaolagoitia:
@Jade-JadeH I tested Pau's branch with 1.3. The only issues I got were:
* in `sample_project\Source\sample_project\Routing\RouteManager.cpp` where the deprecated `GetAPIkey` was being used. We need to use `GetAPIKey`.
* a similar one in a blueprint..

both easily fixable.. I wasn't able to push the fixes probably because this is in Pau's account repo.

Jade-JadeH:
@airaolagoitia Ok thanks was able to get that to work, still have other two issues @PauFlorenti which is that 

1. when you open the flightsim level, and put API key in the ArcGISMapActor, you get the "Game Instance has no valid widget menu!" error and plane just crashed. This error doesn't occur when you open from the sample viewer level. 
2. the H key doesn't work 

Other than this the sample looks great. I do think that when the speed drops the plane's falling speed needs to be faster  @avr757 also the UI needs to be reworked too

PR #48, opened 2023-07-24
Mas12517/user interface/samples UI

mmgaw113:
**Sample**

**NOTE: Make sure you build the solution file**

- Updates to Sample UI
- Add "ArcGISSamples" to the `PublicDependencyModuleNames.AddRange()` to fix a casting issue when casting to the ArcGISPawn.

**Summary**

- Redo Sample UI for Geocoding Sample
- Redo Sample UI for Measure Sample
- Redo Sample UI for Routing Sample

**Tests**

Windows

**ArcGIS Maps SDK Version**

Unreal 5.0.3 w/Maps SDK 1.2


sample_project/Source/sample_project/Measure/Measure.h
```diff
@@ -59,6 +59,19 @@ class AMeasure : public AActor
 	UFUNCTION(BlueprintCallable)
 	void UnitChanged();
 
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+		bool bIsFeet;
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+		bool bIsMeters;
```
dan13436, line 65: Having each unit type as a bool is interesting, wouldn't an enum signifying each unit type be better?

sample_project/Source/sample_project/Routing/RouteManager.cpp
```diff
@@ -173,6 +174,28 @@ void ARouteManager::AddStop()
 	}
 }
 
+void ARouteManager::ClearMap()
+{
+	// Remove the old route cue
+	for (auto SPC : SplineMeshComponents)
+	{
+		SPC->UnregisterComponent();
+		SPC->DestroyComponent();
```
dan13436, line 183: Is it okay to destroy components in an array while iterating through that array

dan13436:
Other than the casting issue in the beginning that we resolved, I did not come across any other bugs

dan13436:
The instructions widget for the geocoding sample is smaller and feels more squished than the instructions widget for the measure and routing sample:
image
image

dan13436:
It is alittle difficult to select different unit types for the measure sample, I would change the button collider to have a bit of margin so you do not always have to be hovering exactly over the text to switch to it
image

dan13436:
> 

The border represents a perfect area of margin in my opinion

dan13436:
It also feels a bit disconnected to have the info toggle button in the top right, while having the actual info widget appear in the bottom right. I would have them be in the same corner for proximity, but that might just be me

sample_project/Source/sample_project/Routing/RouteManager.cpp
mmgaw113: It should be fine since it is destroying the component and simultaneously removing it from the array. It only loops through it once and then empties the array at the end so it should never return a null reference error.

avr757:
> It also feels a bit disconnected to have the info toggle button in the top right, while having the actual info widget appear in the bottom right. I would have them be in the same corner for proximity, but that might just be me

I'll echo to this as well. Besides the feature layer or stream layer where they both have a drop down menu, other samples can have the instruction menu set up like the ones in Figma

image

ZackAllen (approved):

avr757 (approved):
Looking great now! I'll go ahead and approve it but before you merge please take a final check at all the samples to ensure consistency:

1) The cancel button sizes varies from sample to sample. For example, the one from the line of sight sample somehow looks bigger than the routing one. Also the position is different. I think the one from line of sight the size is good, but position needs to be moved further to the right. The position of the cancel button from routing and other samples is good, but the size is a little too small. So I would check all the samples and make sure the size and position are consistent on the cancel button.
image

2) The other thing is the size of the rest button from routing and measure sample. You'll see the reset button and cancel button are in different sizes. They should be the same for consistency reasons. You can decrease the reset button size just a tiny bit and increase the cancel button as mentioned from the 1st comment.
image

3) I'm still getting the error/exception when switching unit randomly in measure samples, the probability of reproducing it is pretty high. Have you tried more testing on that? We should log this issue to investigate further in the future.

4) For the Geocoding sample, can you move the word "intended" or "intended one" to fill the gap? We just need to make sure the margin spaces are equal on the left and right side. Maybe "intended" alone is good enough for the margin spaces to look identical on both sides, but you can try it out and see.
image

5) For the third person controller sample, after pressing "G" to show curse and click on the info icon, I wasn't able to bring out the instructions menu.
UnrealEditor_2HTb8yu3ud

6) The instructions menu for the line of sight sample should be moved a little to the right - to be more consistent with instruction menu positions from other samples.

PR #50, opened 2023-08-21
Mas12517/feature layer redesign

mmgaw113:
**Sample**

Feature Layer Redesign

**Summary**

This sample aids to add functionality to the Original Feature Layer Sample that allows users to customize the data being queried. 

The link provided in the sample is [here](https://services.arcgis.com/V6ZHFr6zdgNZuVG0/ArcGIS/rest/services/ParkTrees/FeatureServer/0/Query?f=geojson&where=1=1&outfields=*)

You may swap the link out for the [baseball stadiums](https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/Major_League_Baseball_Stadiums/FeatureServer/0/Query?f=geojson&where=1=1&outfields=*) (or any other feature layer) at any time

- Please test valid links (put something like https://google.com in and it should give you a warning)
- When you change to a different feature layer, don't change the outfields and it should give a warning
- Uncheck the "Get All Samples" box but leave the index at 0 - 175 and it should give a warning

**Feature Layer Blueprint**
- This blueprint is called bp_feature and is already in the scene. 
- If you click on it you should see this 
image

- Clicking on outfields will show the outfields the user is getting. You may change these at anytime to valid outfields associated with the feature layer you are querying. If you would like to see what the outfields are on the feature layer, go to your feature layer in your web browser and set the outfields to **"outfields=*"** and this will query all of the outfields.

**Objects Spawned In**
- If you set the outfields to valid outfields, you can click on the objects and there will be an array called "Feature Data"
- If you click the array to open it, it should show the data it got from the feature layer query

**Tests**

Windows

**ArcGIS Maps SDK Version**

UE 5.0 with Maps SDK 1.2 & 1.3


sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -57,18 +53,38 @@ class SAMPLE_PROJECT_API AFeatureLayer : public AActor
...
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	FWebLink WebLink;
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	TArray<FProperties> FeatureData;
+	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
+	AArcGISPawn* ArcGisPawn;
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	bool bGetAllFeatures = true;
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	int startValue;
```
NaumanIsGISMan, line 66:
```suggestion
	int StartValue;
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -57,18 +53,38 @@ class SAMPLE_PROJECT_API AFeatureLayer : public AActor
...
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	int lastValue;
```
NaumanIsGISMan, line 68:
```suggestion
	int LastValue;
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -57,18 +53,38 @@ class SAMPLE_PROJECT_API AFeatureLayer : public AActor
...
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	bool bButtonActive;
 	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
-		FFeature data;
+	bool bLinkReturnError;
```
NaumanIsGISMan, line 72: I would group the bools all together

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -57,18 +53,38 @@ class SAMPLE_PROJECT_API AFeatureLayer : public AActor
...
 	virtual void BeginPlay() override;
+	
 public:	
-	// Called every frame
+	UFUNCTION(BlueprintCallable)
+	void ProcessWebRequest();
+	UFUNCTION(BlueprintCallable)
+	bool ErrorCheck();
+	UFUNCTION(BlueprintCallable)
+	void CreateLink();
```
NaumanIsGISMan, line 88: Move and consolidate this public section with the one above, alphabetical, functions first

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -57,18 +53,38 @@ class SAMPLE_PROJECT_API AFeatureLayer : public AActor
 	GENERATED_BODY()
 	
 public:	
-	// Sets default values for this actor's properties
 	AFeatureLayer();
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
```
NaumanIsGISMan, line 57: Do these properties need to be public? Looks like you can make a lot of them private

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -57,18 +53,38 @@ class SAMPLE_PROJECT_API AFeatureLayer : public AActor
...
+	FWebLink WebLink;
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	TArray<FProperties> FeatureData;
+	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
+	AArcGISPawn* ArcGisPawn;
```
NaumanIsGISMan, line 62:
```suggestion
	AArcGISPawn* ArcGISPawn;
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -21,34 +21,30 @@
...
-	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
-	TArray<FString>  TEAM;
-	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
-	TArray<FString>  NAME;
-	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
-	TArray<double>  longitude;
-	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
-	TArray<double>  latitude;
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	TArray<FString> featureProperties;
```
NaumanIsGISMan, line 45:
```suggestion
	TArray<FString> FeatureProperties;
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -21,34 +21,30 @@
...
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	TArray<float> geoProperties;
```
NaumanIsGISMan, line 47:
```suggestion
	TArray<float> GeoProperties;
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -21,34 +21,30 @@
...
-	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	TArray<FString> outFields;
+	FString headers;
 	FString outFieldHeader;
 };
 
 USTRUCT(BlueprintType)
-struct SAMPLE_PROJECT_API FFeature
+struct SAMPLE_PROJECT_API FProperties
```
NaumanIsGISMan, line 40: `FProperties` might be too generic.
```suggestion
struct SAMPLE_PROJECT_API FFeatureLayerProperties
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
```diff
@@ -21,34 +21,30 @@
...
-	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
 	FString link;
-	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
 	FString requestHeaders;
-	UPROPERTY(BlueprintReadWrite, VisibleAnywhere)
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	TArray<FString> outFields;
+	FString headers;
```
NaumanIsGISMan, line 35: `UpperCamelCase` for these public props

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -27,42 +27,95 @@ AFeatureLayer::AFeatureLayer()
...
+	FeatureData.Empty();
 	FHttpRequestRef Request = FHttpModule::Get().CreateRequest();
 	Request->OnProcessRequestComplete().BindUObject(this, &AFeatureLayer::OnResponseRecieved);
-	Request->SetURL(url->link);
+	Request->SetURL(WebLink.link);
 	Request->SetVerb("Get");
 	Request->ProcessRequest();
 }
+
+// Called when the game starts or when spawned
```
NaumanIsGISMan, line 115:
```suggestion
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -27,42 +27,95 @@ AFeatureLayer::AFeatureLayer()
...
+	}
+}
+
+bool AFeatureLayer::ErrorCheck()
+{
+	for (auto Element : FeatureData)
+	{
+		FProperties feature = Element;
+		
+		for (auto property : feature.featureProperties)
```
NaumanIsGISMan, line 70: The naming here is odd. Can you consolidate this loop? Is `Element` the same type as `FProperties`?
```
for (auto data : FeatureData)
{
	for (auto property : data.featureProperties)
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -27,42 +27,95 @@ AFeatureLayer::AFeatureLayer()
...
-			data.latitude.Add(coordinates[1]->AsNumber());
+			TArray<TSharedPtr<FJsonValue>> Features = ResponseObj->GetArrayField("features");
+
+			for (int i = 0; i != Features.Num(); i++)
+			{
+				FProperties testProperties;
+				TSharedPtr<FJsonObject> feature = Features[i]->AsObject();
+				TSharedPtr<FJsonObject> properties = feature->GetObjectField("properties");
+			
+				for(auto outfield : WebLink.outFields)
```
NaumanIsGISMan, line 47:
```suggestion
				for (auto outfield : WebLink.outFields)
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
NaumanIsGISMan: Also looks like these could be `const auto`

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -27,42 +27,95 @@ AFeatureLayer::AFeatureLayer()
...
-			data.TEAM.Add(properties->GetStringField("TEAM"));
-			data.longitude.Add(coordinates[0]->AsNumber());
-			data.latitude.Add(coordinates[1]->AsNumber());
+			TArray<TSharedPtr<FJsonValue>> Features = ResponseObj->GetArrayField("features");
+
+			for (int i = 0; i != Features.Num(); i++)
+			{
+				FProperties testProperties;
+				TSharedPtr<FJsonObject> feature = Features[i]->AsObject();
+				TSharedPtr<FJsonObject> properties = feature->GetObjectField("properties");
```
NaumanIsGISMan, line 45: can be autos? const autos?

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -27,42 +27,95 @@ AFeatureLayer::AFeatureLayer()
...
-	TSharedRef<TJsonReader<>> Reader = TJsonReaderFactory<>::Create(ResponseBody);
-	if (FJsonSerializer::Deserialize(Reader, ResponseObj))
+	if(bConnectedSucessfully)
 	{
-		TArray<TSharedPtr<FJsonValue>> Features = ResponseObj->GetArrayField("features");
-		for (int i = 0; i != Features.Num(); i++)
+		bLinkReturnError = false;
+		TSharedPtr<FJsonObject> ResponseObj;
+		const FString ResponseBody = Response->GetContentAsString();
+		TSharedRef<TJsonReader<>> Reader = TJsonReaderFactory<>::Create(ResponseBody);
```
NaumanIsGISMan, line 35: auto?

sample_project/Source/sample_project/Feature Layer/FeatureLayer.h
mmgaw113: Yes, Blueprint read write properties can't be private. But some aren't edit anywhere so I'll move them around

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
mmgaw113: I don't know if these can be consolidated... The first loop is getting each Feature in Feature Data. Then the second loop is getting the array in the struct of FeatureProperties. So I am not sure if consolidating is possible. Tips or pointers if you know of a way would be much appreciated.

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
mmgaw113: Line 33 requires initialization if using auto

Jade-JadeH:
1. I tried this one https://runtime.maps.arcgis.com/home/item.html?id=d957997ccee7408287a963600a77f61f with this URL https://services9.arcgis.com/RHVPKKiFTONKtxq3/arcgis/rest/services/USA_Wildfires_v1/FeatureServer and, and also this one https://runtime.maps.arcgis.com/home/item.html?id=248e7b5827a34b248647afb012c58787 with this URL https://services9.arcgis.com/RHVPKKiFTONKtxq3/arcgis/rest/services/Active_Hurricanes_v1/FeatureServer and didn't work.

2. When I enter the service the map moves 

3. Can you add a loading indicator and error indicator in the note banner like this one? Or, you could change "Process Request" to "loading" when user inputs a data

image

4. there's still this instruction here, maybe you plan to fix it on another PR: 
image

5. there's no index filtering compared with this PR https://devtopia.esri.com/runtime/normandy/issues/2161#issuecomment-4283144

Nitpicking:
1. The use of pin doesn't seem to make as much sense as routing and geolocating samples @avr757 any ideas?

Jade-JadeH:
Testing this feature service https://runtime.maps.arcgis.com/home/item.html?id=d957997ccee7408287a963600a77f61f with the URL https://services9.arcgis.com/RHVPKKiFTONKtxq3/arcgis/rest/services/USA_Wildfires_v1/FeatureServer doesn't work, tested a couple others and don't work either

mmgaw113:
> Testing this feature service https://runtime.maps.arcgis.com/home/item.html?id=d957997ccee7408287a963600a77f61f with the URL https://services9.arcgis.com/RHVPKKiFTONKtxq3/arcgis/rest/services/USA_Wildfires_v1/FeatureServer doesn't work, tested a couple others and don't work either (个_个)

Add “/query?” To the link and it should work. I’ll add it to the code but it currently assumes you will add “query?”

mmgaw113:
The UI we plan to address in a future PR, this is just for functionality because it was such a massive task/issue, we wanted to split the UI apart from it. Right now, I just want to make sure everything is working properly and then I will disable the old UI before it gets merged in. 

I don't know if anything else makes sense other than a pin at the moment. We changed the default link out to be trees instead of baseball stadiums but if we make the static mesh a tree, then users might not know how to change it to be what they want. So I think leaving it as a pin makes the most sense for now.

mmgaw113:
> 1. When I enter the service the map moves

So I would need to see what you are talking about to understand but there are two things I can think of. Pasting the link into the search bar by hitting "ctrl + v" moves the camera downward. crtl is one of the input mappings for the arcgis Pawn and moves the camera down. 

Alternatively, when you process the request, it randomly picks a point from the feature layer and moves the camera to it.

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -27,42 +27,99 @@ AFeatureLayer::AFeatureLayer()
...
-	UWebLink* url = NewObject<UWebLink>();
-	url->link = "https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/Major_League_Baseball_Stadiums/FeatureServer/0";
-	url->requestHeaders = "f=geojson&where=1=1";
-	url->outFieldHeader = "outfields=TEAM,NAME,LEAGUE";
+	if(!WebLink.Link.Contains("/0/query?"))
+	{
+		WebLink.Link += "/0/query?";
+		bButtonActive = true;
+	}
+	else if(!WebLink.Link.Contains("f=geojson&where=1=1"))
```
Jade-JadeH, line 88:
```suggestion
	else if(!WebLink.Link.Contains("f=geojson&where=1=1&outfields=*"))
```

dan13436 (approved):
I am not too knowledgeable on feature layers yet, but from what I could gather it seemed like it was working. The lack of UI feedback on top of my ignorance did make it hard to manage, but it seems like the functionality is all there. Concatenating the "/0/query?" through code helped a lot; seems to work with Jades link

dan13436:
Would definitely hold off on other approvals before merging, I could have missed something

avr757:
> 1. The use of pin doesn't seem to make as much sense as routing and geolocating samples @avr757 any ideas?

I think the pin is generic enough for now, but if we want to make it more fun and entertaining in the future, we could use lollipop models instead.

image

Jade-JadeH:
so the URL still doesn't work for me, the concatenation seems to add additional outfields:
image

mmgaw113:
> so the URL still doesn't work for me, the concatenation seems to add additional outfields: image

what are you pasting in for the URL?

Jade-JadeH:
> bout to understand but there are two things I can think of. Pasting the link into the search bar by hitting "ctrl + v" moves the camera downward

Both of them. But sounds good that you'll fix it during UI.

mmgaw113:
@Jade-JadeH I swapped the concatenation to a foreach loop. The loop will return invalid link if it does not end in outfields=*. The link will not work if it is not in the correct order of /0/query?f=geojson&where=1=1&outfields=*
I just set it up so that it returns an error now if this happens and will ask you to fix your link.

It should be good for you to take a look again

Jade-JadeH:
I tried this page https://runtime.maps.arcgis.com/home/item.html?id=9e2f2b544c954fda9cd13b7f3e6eebce (with this URL https://services9.arcgis.com/RHVPKKiFTONKtxq3/arcgis/rest/services/USGS_Seismic_Data_v1/FeatureServer).

I got different results which are the follow:

1. Data is not aligned with the terrain

I had this before when I worked on the feature layer sample. This is due to mesh being spawned before terrain is loaded. 

image

2. an empty warning and no data is displayed:

image

3. Data is loaded 

In this case, I noticed that we just zoom to a data point, but instead I think we should "zoom to layer" otherwise it's really hard to get an overview of the data.

Jade-JadeH:
I also tried this page https://runtime.maps.arcgis.com/home/item.html?id=0ec8512ad21e4bb987d7e848d14e7e24 and unreal crashed.

mmgaw113:
> I also tried this page https://runtime.maps.arcgis.com/home/item.html?id=0ec8512ad21e4bb987d7e848d14e7e24 and https://runtime.maps.arcgis.com/home/item.html?id=a98fd08751a5480c898b7cebe38807f4 (not with these two URLs, with the URLs inside the pages)
> 
> and it gives me a warning "URL is not valid" and unreal crashes.

This link isn't going to work because it isn't in the same format as the others. Currently the sample only supports specific types of feature layers, I forget the type but @ZackAllen and I talked a bit about it. We want to support more types in the future but this is already a big enough task as is so we will split it up into smaller tasks once we get both unity and unreal primary functionality working.

Jade-JadeH:
> This link isn't going to work because it isn't in the same format as the others. Currently the sample only supports specific types of feature layers, I forget the type but @ZackAllen and I talked a bit about it. We want to support more types in the future but this is already a big enough task as is so we will split it up into smaller tasks once we get both unity and unreal primary functionality working.

Can you add which data is not supported? And prevent crashes?

mmgaw113:
> Can you add which data is not supported? And prevent crashes?

Crashes I should be able to avoid. I think we should put what data types are supported in the instructions when we redo the UI.
It only supports Point Feature Layer Data I believe. The ones you were trying are polygon

mmgaw113:
Crashes are fixed, the error notification also goes away now if there are no errors. It also checks the type of data now in the script so it should give an error if the data type is unsupported.

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -27,42 +27,113 @@ AFeatureLayer::AFeatureLayer()
...
+	{
+		bLinkReturnError = true;
+		return false;
+	}
+	
+	for (const auto Data : FeatureData)
+	{
+		FFeatureLayerProperties feature = Data;
+		
+		for (const auto property : feature.FeatureProperties)
```
NaumanIsGISMan, line 88: You shouldn't have to need to create a separate local var

```suggestion
	for (const auto feature : FeatureData)
	{		
		for (const auto property : feature.FeatureProperties)
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -27,42 +27,113 @@ AFeatureLayer::AFeatureLayer()
...
+				FFeatureLayerProperties testProperties;
+				auto feature = Features[i]->AsObject();
+				auto properties = feature->GetObjectField("properties");
+			
+				for (auto outfield : WebLink.OutFields)
+				{
+					testProperties.FeatureProperties.Add(feature->GetObjectField("properties")->GetStringField(outfield));
+				}
+				auto type = feature->GetObjectField("geometry")->GetStringField("type");
+				TArray<TSharedPtr<FJsonValue>> coordinates = feature->GetObjectField("geometry")->GetArrayField("coordinates");
```
NaumanIsGISMan, line 52: Can this be auto?
```suggestion
				auto coordinates = feature->GetObjectField("geometry")->GetArrayField("coordinates");
```

NaumanIsGISMan (approved):
Looks good code wise beside some minor nits

ZackAllen (changes requested):
Tested with a feature layer from AGOL and had working results.

Request:
- The current instructions image is invalid in the new workflow. We should delete it and replace it with something correct until the new UI is ready.

Suggestions:
- This code is pretty important, I think we should add more comments explaining how features are parsed from the response.
- Add debug message for unsupported geometry. (Instructions should also mention only point geometry is currently supported)

ZackAllen (approved):

PR #52, opened 2023-08-29
Mas12517/feature layer UI

mmgaw113:
**Sample**

Redesign Feature Layer UI after Improvements

**Summary**

Redesign the Feature Layer UI after the improvements to allow users to customize the data they are querying. 

**Tests**

Windows

**ArcGIS Maps SDK Version**

UE 5.0 w/ Maps SDK 1.3


Jade-JadeH:
I can see that the notification works as expected now. Thanks for working on that. 

1. I tested this data https://runtime.maps.arcgis.com/home/item.html?id=dece90af1a0242dcbf0ca36d30276aa3 with this URL https://services9.arcgis.com/RHVPKKiFTONKtxq3/arcgis/rest/services/Satellite_VIIRS_Thermal_Hotspots_and_Fire_Activity/FeatureServer which is a point layer but doesn't work? Also this one doesn't work either https://runtime.maps.arcgis.com/home/item.html?id=cb1886ff0a9d4156ba4d2fadd7e8a139

2. The ctrl problem is not resolved, any plan on resolving that? 

3. Grammar on instruction: point layer data sets are currently the only data set support ---> supported. Or better, change to: Currently, this sample only supports point layer data. Also delete the "we apologize..." part. Also you should put a space in between two paragraphs. 

4. Change "warning" to "Note", we use that for other samples so that we don't need to change between "instruction" and "warnings". And we should also add "The web link you entered is not valid or not supported" instead of "The web link you entered is not valid"

@avr757 you're going to design the outfields UI right?

mmgaw113:
I will need to test those two feature layers. If they aren’t in the same format is what is being queried they won’t work. 

The control issue is not something I can fix. It needs to be changed in the plug-in, not in the samples. 

Grammar I will correct. 

The reason it is warning instead of note is because the sample won’t work if any of those warnings are active. 

In terms of the ui for the outfields, I am using the same ui designed for the baseball stadium drop down list, it is just in a different location.

avr757:
> @avr757 you're going to design the outfields UI right?

Yes, I'll have the redesign ready some time tmr. It will be very much the same with what Mason currently has now regarding the outfields drop-down. But just to kinda polish up a bit all together with other menus to improve UX and visuals.

mmgaw113:
@Jade-JadeH I just tested both links you provided that you tested yesterday and they both work. There is a problem with the outfields code but that is unrelated to the links.

mmgaw113:
also the feature layers you are testing have 5000 features in them, so having "Get All Features" enabled can freeze the engine if your computer can't handle it. It happened to me every time

mmgaw113:
> 4\. And we should also add "The web link you entered is not valid or not supported" instead of "The web link you entered is not valid

There is a separate warning for unsupported links. That warning only pops up if it detects that it is the wrong type of feature layer. The link warning pops up when it fails to connect to the server. 
image

ZackAllen:
UnrealEditor_1nV8GKt0i6
Default size for markers overlap a lot. It looks bad, marker should be smaller.
We need to update the readme document to reflect the updates to the sample.

avr757:
Some UI visual feedback:

image


1. The entire content is shifted left, it needs to be centered.
2. Bold some of the title texts
3. Extend the service URL input field so it lines up with the min value input box
4. Note has some typos
5. It will look a little better if the Note has a little more negative space under the text.

avr757:
Let's also just get a little bit of more negative space on the bottom and top.

image

ZackAllen (approved):
bp_feature blueprint could use organized sections.

Dynamic scaling looks good!

Jade-JadeH:
still not able to open this https://runtime.maps.arcgis.com/home/item.html?id=cb1886ff0a9d4156ba4d2fadd7e8a139


https://github.com/Esri/arcgis-maps-sdk-unreal-engine-samples/assets/86364673/599af7f5-0812-4551-8a5b-c909e7ce4cbb

avr757:
image

1. Can you decrease the drop-down title size "select/unselect outfields" by 2 pts?
2. The margin of each item selection in the drop-down needs to be increased just by a little bit
3. I see some text got cut off at the bottom of the dropdown
4. Can you change the dropdown arrow icon to match with the design?
5. The reset icon and info icon should be around the same size. The info icon is slightly bigger

Nice reset animation by the way!!

____
Some work from previous comments still needs to be taken care of:

image

image

1. The entire content is shifted left, it needs to be centered.
2. Bold all title texts and down size it by 1 pt or 2
3. Extend the service URL input field so it lines up with the min value input box
4. Note has some typos
5. It will look a little better if the Note has a little more negative space under the text.
6. Make the warning sign icon bigger ( to match up with the sign size in Note)

Be aware of the margin space on the bottom of the menu. 

image

Also, the drop down title "select/unselect outfield(s)" should line up with "service URL" title

image

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -38,51 +38,93 @@ void AFeatureLayer::OnResponseRecieved(FHttpRequestPtr Request, FHttpResponsePtr
...
-				auto properties = feature->GetObjectField("properties");
 
-				/*outfield can be set in the scene on bp_feature
-				this loop will take each outfield set in the scene and check to see if the outfield exists
-				if it does exist, it will return the result of the outfield associated with this feature
-				if it does not exist, it will return and error message in the scene*/
-				for (auto outfield : WebLink.OutFields)
+			if(!bNewLink)
+			{
+				//parse through the features in order to get individual properties associated with the features
```
NaumanIsGISMan, line 44: Just a style nit, we probably should do `// Something something`, so space after comment and first work capitalize

```suggestion
				// Parse through the features in order to get individual properties associated with the features
```

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -38,51 +38,93 @@ void AFeatureLayer::OnResponseRecieved(FHttpRequestPtr Request, FHttpResponsePtr
...
+					auto feature = Features[i]->AsObject();
+					//get the properties field associated with the individual feature
 
-				//To avoid crashes, this checks to see if the type of feature is Point, if so it will get the geometry
-				//if not, it will return an error
-				//current the only type of data supported by this sample is Point Layers, but more will be added in the future.
-				if(type.ToLower() == "point")
-				{
-					for (const auto Coordinate : coordinates)
+					/*outfield can be set in the scene on bp_feature
```
NaumanIsGISMan, line 52: Should probably replace old style C comments `/**/` with `//`

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -38,51 +38,93 @@ void AFeatureLayer::OnResponseRecieved(FHttpRequestPtr Request, FHttpResponsePtr
...
-					bCoordinatesErrorReturn = false;
+					//this will get the type of feature
+					auto type = feature->GetObjectField("geometry")->GetStringField("type");
+					//this will get the geometry or coordinates of the feature
+					auto coordinates = feature->GetObjectField("geometry")->GetArrayField("coordinates");
+
+					//To avoid crashes, this checks to see if the type of feature is Point, if so it will get the geometry
+					//if not, it will return an error
+					//current the only type of data supported by this sample is Point Layers, but more will be added in the future.
+					if(type.ToLower() == "point")
```
NaumanIsGISMan, line 94: Space after `if`, same for any other occurence
```suggestion
					if (type.ToLower() == "point")
```

mmgaw113:
> 5. The reset icon and info icon should be around the same size. The info icon is slightly bigger

The reset icon is the size it was imported in as 29x28, if I make it any bigger it will get blurry and pixelated.

avr757:
> > 5. The reset icon and info icon should be around the same size. The info icon is slightly bigger
> 
> The reset icon is the size it was imported in as 29x28, if I make it any bigger it will get blurry and pixelated.

You can export it at a higher resolution. There is 2x, 3x, and 4x options in Figma. Let me know if that works, or I can get one for you.

mmgaw113:
> You can export it at a higher resolution. There is 2x, 3x, and 4x options in Figma. Let me know if that works, or I can get one for you.

I scaled the info icon down to be the size it is in figma. I had it larger than it is in figma so they are now the same size.

avr757:
Another way easier you can do is just to decrease the info icon. No need to reexport and reimport the rest icon

avr757:
Untitled-2

Looking really nice now! Just a few more things,

1. Let's actually bold the drop-down title "select/unselect outfields", it might look better.
2. The drop-down menu items looks a bit too tight. We should expand the height for users to select those easily. The one in the design is about a good height
image
3. Correct the typos in the Note menu

mmgaw113:
> 3. Also, is there a way to make the text in each item slightly darker? like the one down below. to differentiate from the title text.

I don't think so. The way it works in unreal is you create an item widget and then in the list view set the items to populate the list as that item you created. You can change the scale of the list view but that doesn't effect the items. Just how many are visible on the screen.

mmgaw113:
I can make the text one or two points smaller but that would be the best alternative.

mmgaw113:
I take that back, changing the size of the text determines the size of each item

avr757 (approved):
Looking good! nice work!

sample_project/Content/SampleViewer/Samples/FeatureLayer/readme.md
```diff
@@ -1,23 +1,34 @@
 # Query a feature layer
 
-Query objects from a feature layer and show them on the map.
+Query objects from the feature layer provided, or provide your own, in order to show them on the map.
 
 ![Image of Feature Layer Sample](FeatureLayer.jpg)
```
avr757, line 5:
```suggestion
![Image of Feature Layer Sample](FeatureLayer.png)
```

avr757 (approved):

Jade-JadeH:
There're a lot of improvements! I understand that you want to merge ASAP and this PR is getting a lot of back and forth. however, this is a public sample and we need to be very responsible of what we publish to the public. We don't need to make it perfect or polished, but at least there shouldn't be any commonly-encountered bugs. I'm perfectly ok with merging it in to a dev branch (and I think we're getting close to merge into main), but I think there're still a couple of bugs

1. I didn't look at the code, did you implement the zoom to dataset feature? I see it sometimes zoom in but not really to the correct extent? For example in 23'' it zoom in but the data extent is actually bigger? Correct me if I'm wrong!  

https://github.com/Esri/arcgis-maps-sdk-unreal-engine-samples/assets/86364673/8a9fb6c0-edbe-477d-9bf8-fd468ff9ab97

2. if you put edge cases such as 0-0 we have a jittering effects here but instead we should show nothing. 

image

3. There's an issue with scaling but I'm ok with improving this in the future PRs. 

The issue is that when you zoom in further, they're either too sparse or too dense. Looking at other Esri products you'll see the difference. 

In this example of having 6 features, the markers look like 2

image

In this example of having all features, they still look cluttered. 

image

sample_project/Content/SampleViewer/Samples/FeatureLayer/readme.md
mmgaw113: I downloaded the image from you and then converted it to a jpg, this change would break the readme

sample_project/Content/SampleViewer/Samples/FeatureLayer/readme.md
```diff
@@ -1,23 +1,34 @@
...
 
-## How to use the sample
+## How to Setup (Feature Layer Level)
 
 1. Open the **FeatureLayer_lvl** level.
 2. Click on the **ArcGISMapActor** in the Outliner panel.
 3. Set your API key under Authentication section in the Details panel.
 4. Click play and select a stadium from the list.
```
Jade-JadeH, line 12: This part is outdated.

mmgaw113:
> 3\. if you put edge cases such as 0-0 we have a jittering effects here but instead we should show nothing.

putting in 0, 0 should only get the first value of the array. Same goes for putting in the same min and max at any value. I will look into this bug.

sample_project/Content/SampleViewer/Samples/FeatureLayer/readme.md
```diff
@@ -1,23 +1,34 @@
...
 
+## How to Setup (Sample Viewer)
+
+1. Click Play in Unreal Editor.
+2. Input your API key under the **API Key Drop down**.
+3. Click the **Sample Drop Down** and select **Feature Layer**.
+
+## How to Use
+
```
Jade-JadeH, line 21: You should mention that we only support point feature layer here as well

Jade-JadeH:
> > 1. I didn't look at the code, did you implement the zoom to dataset feature? I see it sometimes zoom in but not really to the correct extent? For example in 23'' it zoom in but the data extent is actually bigger? Correct me if I'm wrong!
> 
> No I didn't. I spent all day on it yesterday and couldn't make any progress because the functions I need are not exposed for me to access. It was causing a lot of problems. I raised the altitude it moves the camera to, however for some of the feature layers there is no way to encapsulate all of the data in the screen with out zooming out to the maximum (see the slack image I sent earlier)
> 
> I reached out to Andres to see if there is a way to create a layer struct based off the feature data and in return make it a layer we can zoom to but I don't believe it is possible. So there will likely need to be some sort of compromise on this because I don't think it is possible at the moment.

Ok, looks like this would require maybe a week of investigation. I'm ok with having it as it is right now and mark as a future improvements. Have you look at how we achieve zoom to layer in the modes panel?

mmgaw113:
> The issue is that when you zoom in further, they're either too sparse or too dense. Looking at other Esri products you'll see the difference.
> 
> In this example of having 6 features, the markers look like 2

There isn't much we can do to fix this, it is what happens when you scale up and down based on distance. The only way to make it less of a problem is either not scaling them and keeping them (which makes it so you can't see them at all when zoomed out), or alternatively we move the points which would take away from what the goal of the sample is achieving. 

This issue is present in baseball stadiums in unity. If you look at the LA Dodgers and LA Angels stadiums and zoom out they get closer and closer together. They may not end up touching or overlapping because there is distance between them but this is just how distance based scaling works in this case.

Jade-JadeH:
> > The issue is that when you zoom in further, they're either too sparse or too dense. Looking at other Esri products you'll see the difference.
> > In this example of having 6 features, the markers look like 2
> 
> There isn't much we can do to fix this, it is what happens when you scale up and down based on distance. The only way to make it less of a problem is either not scaling them and keeping them (which makes it so you can't see them at all when zoomed out), or alternatively we move the points which would take away from what the goal of the sample is achieving.
> 
> This issue is present in baseball stadiums in unity. If you look at the LA Dodgers and LA Angels stadiums and zoom out they get closer and closer together. They may not end up touching or overlapping because there is distance between them but this is just how distance based scaling works in this case.

If you tweak the scaling parameters as to how they react to distance I have a sense that you can get a better scaling, it might require some time though

Jade-JadeH:
With this dataset https://runtime.maps.arcgis.com/home/item.html?id=248e7b5827a34b248647afb012c58787

I found that when I put it in for the first time there's the check link warning, on the second time it does show the data. Could you check this?

Jade-JadeH:
I also found that some of the feature services have multiple point layers, https://runtime.maps.arcgis.com/home/item.html?id=248e7b5827a34b248647afb012c58787 Could you make sure that we handle all of them? 

See the preview for the above dataset, different point layers are represented by different point icons:

image

Jade-JadeH:
I see this in the Unity sample. Can you try this mouse over event to see if it resolves the issue of the ctrl camera movement? 

image

mmgaw113:
> Seems like this is a common issue, this needs to be fixed since the purpose of this sample is to let the users use their dataset

This bug is a direct result of getting the outfields when changing the link. It processes the link and gets the outfields but then fails afterwards when trying to actually get the data because it has already processed everything once. 

> I also found that some of the feature services have multiple point layers, https://runtime.maps.arcgis.com/home/item.html?id=248e7b5827a34b248647afb012c58787 Could you make sure that we handle all of them?
> 
> See the preview for the above dataset, different point layers are represented by different point icons:
> 
> image

This will work but the user needs to change the outfields to get the /1/ layer instead of the /0/ layer.

Jade-JadeH:
> > Seems like this is a common issue, this needs to be fixed since the purpose of this sample is to let the users use their dataset
> 
> This bug is a direct result of getting the outfields when changing the link. It processes the link and gets the outfields but then fails afterwards when trying to actually get the data because it has already processed everything once.

I didn't change the link. What I did was enter the link, click process, then the error shows, if I click process again, I can see the data.

Jade-JadeH:
> This will work but the user needs to change the outfields to get the /1/ layer instead of the /0/ layer.

That makes sense, I remember you mentioned appending /0 to the link. I think it will make sense to have a loop to append /1 etc. until it's no longer valid, and use different representations for each layer. This can be future work if not for this PR.

mmgaw113:
> > This will work but the user needs to change the outfields to get the /1/ layer instead of the /0/ layer.
> 
> That makes sense, I remember you mentioned appending /0 to the link. I think it will make sense to have a loop to append /1 etc. until it's no longer valid, and use different representations for each layer. This can be future work if not for this PR.

There is a point where there isn't anything we can do in terms of correcting user specific issues, this is one of them. @ZackAllen and I have had a few discussions in regards to what we should focus on and what we shouldn't and I don't think this is one of them because it is user specific.

We set the default request headers to be:
{
/0/
query?
f=geojson
where=1=1
outfields=*
}

the for loop checks to see if the current index of the array is already at the end of the link, if not it adds it. So at a certain point the user will need to go in and change those for their specific link... Or they can just input the entire link ending in: 
/0/query?f=geojson&where=1=1&outfields=*

or whatever they specifically need it to be. From my research, each feature layer needs to have f=geojson and where=1=1 in order to even work other wise it will return an error in the browser. So I don't think changing how the current functionality works is worth doing unless we receive feedback that explicitly states it is causing issues in which case I think the best action is to guide the user through using it.

Additionally, the documentation, whenever we get around to it, will be explicit enough so that the user understands it. The readme is basic, but I believe we plan to add some extensive instructions since this sample is by far the most complex to construct and use.

mmgaw113:
> I didn't change the link. What I did was enter the link, click process, then the error shows, if I click process again, I can see the data.

Still the same issue, on start it is querying the outfields in order to populate the list. I'm not sure what the best course of action for fixing it is at the moment. I am still looking into it.

Jade-JadeH:
> There is a point where there isn't anything we can do in terms of correcting user specific issues, this is one of them. @ZackAllen and I have had a few discussions in regards to what we should focus on and what we shouldn't and I don't think this is one of them because it is user specific. 

When you open the preview page of this data https://runtime.maps.arcgis.com/home/item.html?id=248e7b5827a34b248647afb012c58787 it does show all layers without having to specify /0 or /1, so I feel like the good UX would be to do this on our end. But at this point we should at least add to the doc.

mmgaw113:
> When you open the preview page of this data https://runtime.maps.arcgis.com/home/item.html?id=248e7b5827a34b248647afb012c58787 it does show all layers without having to specify /0 or /1, so I feel like the good UX would be to do this on our end. But at this point we should at least add to the doc.

image
image
image

Yes, but if you look at the scene viewer, map viewer or Rest Services Directory it shows a list of all the features and which ones are currently being shown. In the Rest Services Directory, in order to query you have to select one. From my understanding, there is no way to get all the features in an HTTP Request with out directly specifying which layer to get (/0/ or /1/ etc).

There have already been numerous scope issues with this sample redesign. Adding the ability to viewer multiple layers at once needs to be a separate project/sample/feature

Jade-JadeH:
> There have already been numerous scope issues with this sample redesign. Adding the ability to viewer multiple layers at once needs to be a separate project/sample/feature
Yeah as I said, this can be a separate future PR if we decided to do it.

mmgaw113:
> With this dataset https://runtime.maps.arcgis.com/home/item.html?id=248e7b5827a34b248647afb012c58787
> 
> I found that when I put it in for the first time there's the check link warning, on the second time it does show the data. Could you check this?
> 
> Also with this dataset as well https://runtime.maps.arcgis.com/home/item.html?id=cb1886ff0a9d4156ba4d2fadd7e8a139
> 
> Seems like this is a common issue, this needs to be fixed since the purpose of this sample is to let the users use their dataset..

I managed to get this one fixed but the solution was running the full query (as opposed to outfields only) on start. This is how the old feature layer was, it would run on start. The user can still input their own data in at any time and modify it prior in bp_feature.

mmgaw113:
> I see this in the Unity sample. does this help with the issue of the ctrl camera movement?

I disabled Input on the player controller which achieves the same effect.

Jade-JadeH:
A lot of previous issues I have seen have been fixed. The UI feels smooth and makes sense to me. The UI animation is nice. Great work! I'm ok with merging since there are no common bugs. These are items I tested:

1. Multiple point layers from living atlas. 
2. Polygon and other unsupported layers show warning correctly 
3. number of features matches entry
4. Check/uncheck get all features
5. Select outfields

There are additional comments and questions but feel free to merge! 

1. I noticed some layers trigger the coordinates not found error, I suspect that they do have coordinates by looking at the preview such as this one: https://runtime.maps.arcgis.com/home/item.html?id=33820e818ebc4661b01bcd47e5f2a57e, just don't follow the format that we defined. I suggest to change the warning message to: coordinates not supported instead of not found. 

2. I noticed that the feature items numbering doesn't start with 0
image

3. What does select/unselect outfields do? It doesn't change feature items shown

Jade-JadeH (approved):

mmgaw113:
> 3\. What does select/unselect outfields do? It doesn't change feature items shown

If you click on an individual feature, for instance bp_featureitem176, it will show an array of feature properties, which are the outfield you select. If you have get all outfields selected, it will get all of them. 
I just checked and there is an issue with it on start so I will deal with that monday before merging.

Jade-JadeH:
- If you click on an individual feature, for instance bp_featureitem176, it will show an array of feature properties, which are the outfield you select. If you have get all outfields selected, it will get all of them.

This makes sense to me. 

- I just checked and there is an issue with it on start so I will deal with that monday before merging.

Oh ok, I think I saw it too, the outfields don't get populated on start. 

image

mmgaw113:
> Oh ok, I think I saw it too, the outfields don't get populated on start.

They do now, it was a setting in the level I didn't push because of API key so it works now and readme is up to date. I had already updated the readme but it got overwritten from Andy's image updates. Merging this now.

PR #53, opened 2023-09-12
Readme cover image update

avr757:
Update readme cover images to all samples.

- Unified cover image sizes to 1280x720 res
- Changed all image format to .png
- Changed readme files


sample_project/Content/SampleViewer/Samples/Measure/readme.md
```diff
@@ -2,7 +2,7 @@
 
 Display the measured distance by drawing lines on the map.
```
ZackAllen, line 3: Metadata.json is not changed for measure, was it already correct?

ZackAllen (changes requested):

Jade-JadeH:
Routing sample one I think the angle could be more top-down like the Unity one. 

Third person one---love it!

sample_project/Content/SampleViewer/Samples/Measure/readme.md
avr757: Yes, the image name is already matching 
image

avr757:
> Routing sample one I think the angle could be more top-down like the Unity one.

How about this one?


Routing

Jade-JadeH:
I think the Unity one's angle looks better, I think it's better to not have a birdview so that it looks completely 2D.

avr757:
> I think the Unity one's angle looks better, I think it's better to not have a birdview so that it looks completely 2D.

How about this angle?

Routing

Jade-JadeH:
Thanks for keep working on it.

Jade-JadeH (approved):

ZackAllen (approved):

PR #54, opened 2023-10-06
Split samples into general samples project and xr samples project

mmgaw113:
**Sample**

Split Samples Project

**Summary**

- Created a new project explicitly for XR samples in Unreal Engine
- Moved GitIgnore to Parent Directory in the project
- Added necessary lines in GitIgnore to ignore the same files in the xr samples project as in the samples project

**How to Test**

- Clone the branch *Mas12517/split project*
- Navigate to repository directory in File Explorer
- Open the folder labeled xr_sample_project
- Right click on xr_samplesproject.uproject and click "Generate Visual Studio Project Files"
- Once complete open git and ensure that there are no new files are being tracked
- Open the project

**Tests**

Windows, will test mac in office

**ArcGIS Maps SDK Version**

N/A & Unreal 5.3.1


NaumanIsGISMan:
Could you rename the PR to something more descriptive, something like `Split repo into samples project and XR project`

NaumanIsGISMan (changes requested):
You can have multiple gitignores at different folder levels instead of one massive gitignore with a lot of repeating patterns. 

Can you break apart this gitignore into three gitignores? One will be at the top level with ignores that are common across the repo (you can revert/restore the one you deleted), one that will go inside the top level of the xr project with its respective ignores and the other at the top level of the main sample repo with its ignores.

Also there's a lot of things that can just stay the same, like the `*.slo` and `*.lo`, you dont necessarily need to have them be project specific, with some exceptions.

NaumanIsGISMan (approved):

ZackAllen (approved):

PR #65, opened 2023-11-23
Mas12517/stream layer UI

mmgaw113:
**Sample**

Stream Layer UI

**Summary**

- Added the UI to the Stream Layer sample
- Fixed a bug where planes were not rotated properly
- Added the camera moving behind the plane

**Tests**

Windows

**ArcGIS Maps SDK Version**

UE 5.0 w/ Maps SDK 1.3


mmgaw113:
> 3\. This is more related to UX. I think the camera movement (WASD) sensitivity should be decreased as what we have in the Unity Stream Layer. The camera moves too fast in this case.

This is based on the ArcGISPawn that is provided with the plugin. It accelerates/decelerates based on where the pawn is in the world. Not much I can do for this aside from remove the Movement Component and make it so the user can't move the plane at all.

avr757 (approved):
looks great and nice job! thanks for fixing those.
just one quick thing to revert the scroll bar color back to white in sample viewer and it's good to go!

image

PR #78, opened 2024-04-25
Add support for Unreal Engine 5.4 with Release of Maps SDK 1.6

mmgaw113:
**Sample**

Sample Project
XR Sample Project

**Summary**

Add Support for Unreal Engine 5.4 in XR Sample Project (with release of 1.6)
Add Support for Unreal Engine 5.4 in Sample Project (with release of 1.6)
Update Sample Project to Unreal Engine 5.2 with deprecation of UE5.0 and 5.1

**Tests**

Windows

**ArcGIS Maps SDK Version**

Maps SDK 1.6 (based off andres's UE 5.4 branch)


sample_project/Source/sample_project/3DAttribute/AttributeComponent.h
```diff
@@ -5,6 +5,7 @@
 //
 #pragma once
 
+#include <functional>
```
airaolagoitia, line 8: this one should be moved to the plugin .. potentially to `ArcGISMapsSDK/BlueprintNodes/GameEngine/Layers/ArcGIS3DObjectSceneLayer.h` or any other header in the include chain

xr_sample_project/Source/xr_samplesproject/VRSample/VRCharacterController.h
```diff
@@ -124,6 +124,8 @@ class XR_SAMPLESPROJECT_API AVRCharacterController : public ACharacter
 	UMotionControllerComponent* rightMotionControllerInteractor;
 	UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Variables", meta = (AllowPrivateAccess = "true"))
 	USpringArmComponent* springArmComponent;
+	UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Variables", meta = (AllowPrivateAccess = "true"))
+	TEnumAsByte<EHMDTrackingOrigin::Type> TrackingOrigin;
```
airaolagoitia, line 128: are we sure that the default value of this property for unreal <= 5.3 represents the same thing as the default value in unreal 5.4. if not, we will have a behavior difference between unreal versions

mmgaw113: Yes, I verified the sample still works and compiles in both 5.3 and 5.4 despite the default values being different. Making it a variable allows users to change the tracking origin type in the editor to keep the "low code - no code" abilities that our samples are intended to provide.

ZackAllen:
fZOFd8iSts
Any way to address this warning?
Should we update `DefaultEngine.ini` or not?

ZackAllen:
Tested main samples for 5.2 and everything works as expected. Still need to verify XR samples.

mmgaw113:
> fZOFd8iSts Any way to address this warning? Should we update `DefaultEngine.ini` or not?

Does this pop up in 5.2 or just 5.4? I know it started showing up in 5.3 but I never used 5.2 really. So we need to verify if it shows up in 5.2 and if it doesn't we need to ensure that enabling this doesn't cause issues in 5.2.

NaumanIsGISMan:
> Does this pop up in 5.2 or just 5.4? I know it started showing up in 5.3 but I never used 5.2 really. So we need to verify if it shows up in 5.2 and if it doesn't we need to ensure that enabling this doesn't cause issues in 5.2.

It pops up in all versions, I've seen it in 5.1, 5.2, and 5.3

xr_sample_project/Source/xr_samplesproject/VRSample/VRCharacterController.h
airaolagoitia: yeah but if users don't change it manually, they will get different results in `=<5.3` and `5.4`

You could do this to address it

```suggestion
#if UE_VERSION_OLDER_THAN(5, 4, 0)
	TEnumAsByte<EHMDTrackingOrigin::Type> TrackingOrigin{EHMDTrackingOrigin::Floor};
#else
	TEnumAsByte<EHMDTrackingOrigin::Type> TrackingOrigin{EHMDTrackingOrigin::LocalFloor};
#endif
```

sample_project/Source/sample_project/3DAttribute/AttributeComponent.h
airaolagoitia: should we use `TFunction<...>` to avoid this std include?

xr_sample_project/Source/xr_samplesproject/VRSample/VRCharacterController.h
mmgaw113: When we update the project to 5.4, you won't be able to open any files, levels, blueprints, etc, in 5.3. So when we go through the process of changing the version the project uses, we will update the level to have this value be the equivalent of what it is in 5.3 (floor = local floor). I don't think we need to add this conditional for something that will be removed in ~ 3 months and since the plugin doesn't work in 5.4 currently, users don't need it now either.

ZackAllen (approved):
Verified on Valve Index in Unreal 5.3. Everything works as expected.

PR #84, opened 2024-06-05
Jgom27/Blueprint sample - Third Person

Jgom27:
**Sample**

Adding a new version of the third person character sample made entirely with blueprints.

**Summary**

This project is a blueprint implementation of the original third person character sample. It replicates the functionality of the original C++ version using Unreal Engine's blueprint system.

**Tests**

Tested with Windows PC, operating system: Windows 10

**ArcGIS Maps SDK Version**

Version 1.5.0 of the ArcGIS Maps SDK for Unreal Engine was used.


ZackAllen (changes requested):
We should not duplicate all of the mapman assets. We should re-use what is already present in the repo.

mmgaw113 (changes requested):
Overall looks good, good job. A couple of things:

- The directions show up to start, but once you hide them, you can't get them to show up again by clicking on the info icon. Our users need to be able to pull them up at any point if they would like. Check out wbp_infobutton, that is where you'll want to add the fix. 
- Can you move the folder you created for the blueprint specific version of the sample into the ThirdPersonCharacter folder and call it something like "blueprint version" and then make a new folder and call it "c++ version" and move the c++ level and character into that folder just to tity things up a bit? Let me know if this breaks anything.

mmgaw113 (approved):
Looks good, nice job!

ZackAllen (approved):

PR #87, opened 2024-07-09
Feature/Building Filter

Jgom27:
**Sample**

Building scene layer sample added: Building Explorer

**Summary**

Added a new sample, Building Explorer, which allows users to explore a building scene layer by toggling the visibility of different disciplines and categories, filtering by building level, and filtering by construction phase. 

**Tests**

Tested on windows 11 pc via unreal engine 5.0.

**ArcGIS Maps SDK Version**

ArcGIS Maps SDK v1.5.0

image


Jade-JadeH:
Is this good for review now or later

Jgom27:
> Is this good for review now or later
Good for review.

Jade-JadeH:
Tested different menu items in BuildingExplorer and work as expected. So good work on the UI part. 

Also found some potential issues

1. When you click from sample viewer level, the UI is incomplete for me, and at one point it also crashes

image

2. The samples in our sample viewer level is supposed to sort alphabetically (although I do notice that the order of geocoding and hit test is wrong too). 

image

3. When there's no API key, the menu still shows some default levels. 
image

4. The eye icon seems pretty blurry. I didn't know whether it's to check or uncheck in the beginning. 
image

5. I believe here's the new D&C order. We changed it in the UI recently. The architectural one is the most used one so it makes sense to put it in the beginning. 
image

I also have a question which is what is the use of the dustbin icon?

sample_project/sample_project.uproject
```diff
@@ -9,14 +9,19 @@
...
 		{
 			"Name": "EnhancedInput",
 			"Enabled": true
+		},
+		{
+			"Name": "ArcGISMapsSDK",
+			"Enabled": true
 		}
 	]
 }
```
Jade-JadeH, line 27: is this change necessary?

avr757:
Nice job overall! Just some quick comments on UI components.

1. I think it would be better to leave a little bit of more margin (top and right) for the menu. Also, the logo size should line up with the title "Building Explorer".  The logo + title should aligned horizontally centered with the menu background.
image

2. There should be more spacing between the Construction Phase and the slider, and the slider should be a little wider as the design. The Construction Phase Number can be moved slightly to the right to align with the Trash Can icon as the design. And as Jade mentioned, the Visibility icons seem distorted. Can you make the Structural, Electrical, Architectural fonts smaller? They should be in the same sizes as the Level or Construction Phase. Lastly, all the fonts under the Disciplines and Categories look different. They should be in "Chakra Petch" type face.
image

3. The trash bin icon should clear the level. 
4. There should be a space between Building and Explorer. And it should be in alphabetical order in the sample view list.
image

mmgaw113 (changes requested):
I already mentioned the ArcGISPawn copy and Blueprint for the c++ class. Also please go through and revert the [changes](https://dev.to/lofiandcode/git-and-github-how-to-revert-a-single-file-dha) to `sample_project.uproject`. Unless we are adding plugins like OpenXR or water plugin, we don't update the Uproject to add dependencies for our plugin.

sample_project/sample_project.uproject
Jgom27: No, accidentally pushed this.... Will revert

Jade-JadeH (changes requested):
see previous

mmgaw113:
I'm going to take a look at it again this morning. Make sure you update the main readme. The link you use for linking to the readme for this sample won't work because it will be merged into main. So you will need to predict the link. Take a look at the others in that file.

mmgaw113:
By default, the level doesn't show. After clicking on it, it shows up.
image

mmgaw113:
cc @avr757 Should we have an `Enable All` and a `Disable All` button within the UI similar to in disciplines and categories?

Jade-JadeH:
Thanks for address the feedback, and also change the order of hit test and geocoding : )

For point 3, I still see the total number of construction phase. This behavior needs to be aligned with arcgis online 
image

For point 4, I see eye icon are much more visible now. But I noticed that the visible eye icon has two shapes which I think should only have one
image

For point 5, I see in unity main we have architectural-structural-electrical whereas in sample the order is different

Jgom27:
> By default, the level doesn't show. After clicking on it, it shows up. Same with construction phase
> 
> image

- The levels not showing is intentional to align with the Javascript sample:

image

- The construction phases not showing is now fixed.

Jgom27:
> Thanks for addressing the feedback, and also changing the order of hit test and geocoding : )
> 
> For point 3, I still see the total number of construction phase. This behavior needs to be aligned with arcgis online image
> 
> For point 4, I see eye icon are much more visible now. But I noticed that the visible eye icon has two shapes which I think should only have one image
> 
> For point 5, I see in unity main we have architectural-structural-electrical whereas in sample the order is different

- Point 3 should be fixed now
- I have changed the order to architectural-structural-electrical. The Javascript sample does use alphabetical order, but I think being in line with the plugin is probably a good idea. @avr757 any thoughts?
- I cannot recreate the visible icon shape discrepancy. This is very odd.

avr757:
@Jgom27 
One last thing, let's make sure there is a little more space between the logo and building exporerer, and align the logo up with level below:

image



So something like this:

image

sample_project/Source/sample_project/BuildingExplorer/SimpleBuildingSceneLayerActor.cpp
```diff
@@ -0,0 +1,340 @@
...
+		if (ArcGISMapActor)
+		{
+			ArcGISMapComponent = ArcGISMapActor->FindComponentByClass<UArcGISMapComponent>();
+			if (ArcGISMapComponent != nullptr)
+			{
+				const auto& AllLayers = ArcGISMapComponent->GetLayers();
+
+				for (const auto& Layer : AllLayers)
+				{
+					if (Layer.Type == 7)
```
ZackAllen, line 71:
```suggestion
					if (Layer.Type == EArcGISLayerType::ArcGISBuildingSceneLayer)
```

sample_project/Source/sample_project/BuildingExplorer/SimpleBuildingSceneLayerActor.cpp
```diff
@@ -0,0 +1,340 @@
...
+		if (loadError)
+		{
+			LoadStatus = TEXT("Failed");
+		}
+		else
+		{
+			LoadStatus = TEXT("Loaded");
+		}
+	});
+	if (LoadStatus == TEXT("Failed"))
```
ZackAllen, line 124: Don't use a string comparison for this, add a boolean in the function.

sample_project/Source/sample_project/BuildingExplorer/SimpleBuildingSceneLayerActor.cpp
```diff
@@ -0,0 +1,340 @@
...
+					}
+					DisciplineCategoryData.Add(NewDiscipline);
+				}
+			}
+		}
+	}
+
+	// Define the order
+	TMap<FString, int32> DisciplineOrder;
+	DisciplineOrder.Add(TEXT("Architectural"), 0);
```
ZackAllen, line 161: Architectural
Structural
Mechanical
Electrical
Piping

sample_project/Source/sample_project/BuildingExplorer/SimpleBuildingSceneLayerActor.cpp
```diff
@@ -0,0 +1,340 @@
...
+	if (!bClearLevel)
+	{
+		WhereClause = FString::Printf(TEXT("%s and %s"), *BuildingLevelClause, *ConstructionPhaseClause);
+	}
+
+	Esri::GameEngine::Layers::BuildingScene::ArcGISBuildingAttributeFilter* LevelFilter;
+	for (auto Filter : Filters)
+	{
+		FString Name = Filter.GetName();
+		if (Name == TEXT("Filter"))
```
ZackAllen, line 228:
```suggestion
		if (Filter.GetName() == TEXT("Filter"))
```

sample_project/Source/sample_project/BuildingExplorer/SimpleBuildingSceneLayerActor.cpp
```diff
@@ -0,0 +1,340 @@
...
+			solidDefinition.SetWhereClause(WhereClause);
+			BuildingSceneLayer->SetActiveBuildingAttributeFilter(Filter);
+		}
+	}
+}
+
+void ASimpleBuildingSceneLayerActor::PopulateSublayerMaps(FString option, bool bVisible)
+{
+	// Search for given discipline/category
+	if (BuildingSceneLayer)
```
ZackAllen, line 241:
```suggestion
	if (!BuildingSceneLayer)
	{
	    return;
	}
```

sample_project/Source/sample_project/BuildingExplorer/SimpleBuildingSceneLayerActor.cpp
```diff
@@ -0,0 +1,340 @@
...
+	{
+		const auto& FirstLayers = BuildingSceneLayer->GetSublayers();
+		for (const auto& FirstSubLayer : FirstLayers)
+		{
+			if (FirstSubLayer.GetName() == TEXT("Full Model"))
```
ZackAllen, line 246: You are enabling all of the full model sublayers, but you never disable the Overview layer.

sample_project/Source/sample_project/BuildingExplorer/SimpleBuildingSceneLayerActor.cpp
```diff
@@ -0,0 +1,340 @@
...
+			ConstructionPhases += TEXT("', '");
+		}
+		else
+		{
+			ConstructionPhases += TEXT("')");
+		}
+	}
+
+	// Create the where clauses
+	FString BuildingLevelClause = FString::Printf(TEXT("BldgLevel in %s"), *BuildingLevels);
```
ZackAllen, line 215: We user `auto` for these.

ZackAllen (changes requested):
Some initial feedback, I'll look more Monday.

sample_project/Source/sample_project/BuildingExplorer/SimpleBuildingSceneLayerActor.cpp
```diff
@@ -0,0 +1,354 @@
+/* Copyright 2023 Esri
```
arash-sh, line 1: I believe we are supposed to use the current year for all new files

arash-sh:
Nice sample! The level and construction phase options don't seem to work when the level is opened directly (without sample viewer).

arash-sh:
> * The levels not showing is intentional to align with the Javascript sample:

I think in that sample, only the level selected is shown (not all levels up to that number). Should the label ("Level") be renamed to make the behavior more clear? 
Also, would it make sense to show the maximum number of levels and the highest construction phase by default?

Jgom27:
> > * The levels not showing is intentional to align with the Javascript sample:
> 
> I think in that sample, only the level selected is shown (not all levels up to that number). Should the label ("Level") be renamed to make the behavior more clear? Also, would it make sense to show the maximum number of levels and the highest construction phase by default?

Went ahead and changed the functionality to only show the level selected. We do show the highest construction phase, but the highest level would not include some of the categories so having a reset button that clears the levels enables a little more functionality. However now, by default, we are showing all the levels so I think this would clear up any confusion the user might have as far as what the trash button would do.

Jgom27:
> Nice sample! The level and construction phase options don't seem to work when the level is opened directly (without sample viewer).

This should be fixed now.... I did not know you could set the api key in the plugin settings.... oops.... ( I was always setting it in the map component)

ZackAllen (approved):

mmgaw113 (approved):
Approving to not block from being merged.

arash-sh:
When adding new bsl, we currently add a new layer. How do you all feel about replacing the old one with the new one instead?
The main disadvantage in the current implementation is that if you add a different model of the same building (e.g., a newer version of building E bsl), both layers will be there and on top of each other.

sample_project/Content/SampleViewer/Samples/BuildingFilter/README.md
```diff
@@ -0,0 +1,44 @@
...
+1. **Open the BuildingFilter Level:**
+   - Navigate to the **BuildingFilter** level in your project and open it.
+2. **Set the API Key:**
+   - In the outliner, locate and select the **ArcGISMapActor**.
+   - In the **Details** panel, find the field for the API key and enter your API key.
+3. **Start the Simulation:**
+   - Click the play button to start the simulation.
+4. **Interact with the UI:**
+   - **Service URL:**
+     - There is a space in the UI where you can add a service URL for a building scene layer. Enter the URL to load a specific building scene layer. You can also use a local filepath for the URL.
```
arash-sh, line 24:
```suggestion
     - There is a space in the UI where you can add a service URL for a building scene layer. Enter the URL to load a specific building scene layer. You can also use a local file path for the URL.
```

sample_project/Content/SampleViewer/Samples/BuildingFilter/README.md
```diff
@@ -0,0 +1,44 @@
+# Building Filter
+
+Allows users to toggle the visibility of different levels, construction phases, disciplines, and categories within a building scene layer. This sample demonstrates how to explore building scene layers and filter based on various criteria.
```
arash-sh, line 3:
```suggestion
This sample demonstrates how to explore building scene layers. User can toggle the visibility of different levels, construction phases, disciplines, and categories within the building scene layer.
```

arash-sh: Not really important but I feel the main thing is exploring bsls

arash-sh (approved):

PR #91, opened 2024-08-19
Arash/xr table top fix

arash-sh:
**Sample**

New XR tabletop sample 

**Summary**

- Scripts for configuring the tabletop and interacting with it
- Scripts for grabbing the table from distance
- Blueprints for UI, loading the preset locations, and table base  
- Misc. assets

**Tests**

N/A

**ArcGIS Maps SDK Version**
1.6


xr_sample_project/xr_samplesproject.uproject
mmgaw113, line 1: Please revert the changes to this file or make sure the branch is up to date with main.

xr_sample_project/Content/Samples/XRTableTop/XRTabletop.png
mmgaw113, line 1: Did you get rid of the Table that was underneath the map?

arash-sh: Yes, based on feedback from Zack, removed to reduce issues with map elevation offset while zooming

mmgaw113: Take a look at XR Table top in unity. I am setting the elevation offset every time the location changes through the UI. Based on feedback from adrien and rex (and maybe users), we should have a table rather than just the map floating. Setting the elevation offset everytime makes it so it is never significantly higher than the table and zooming too far doesn't make the map disappear.

arash-sh: That's how its being done here as well. But after loading a new location, zooming moves the map up and down and can move the map below the table or too far above it. That's because the elevation offset is being multiplied with the map scale, and as far as I know, that's the same behavior in Unity. 
If the table mesh is absolutely needed, one option would be to have the map's base always be at the same height, taking into account the elevation offset but independent of the zoom, which can cause bigger gaps above the table in mountainous areas.
When/if we add an auto offset feature to the tabletop component, this will not be an issue any longer.

mmgaw113: Gotcha, I do remember that being a problem where the map would sometimes go in or below the table.

xr_sample_project/Config/DefaultEngine.ini
arash-sh, line 1: Re-enabled anti-aliasing to avoid flickering on the map and table

xr_sample_project/Source/xr_samplesproject/XRTableTop/XRTabletopComponent.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+bool UXRTabletopComponent::FindArcGISCameraLocationInHierarchy()
+{
+	ArcGISCameraLocation = nullptr;
+	TArray<AActor*> actors;
+	GetOwner()->GetAttachedActors(actors, true, true);
+	for (auto a : actors)
+	{
+		if (a->FindComponentByClass<UArcGISCameraComponent>() && a->FindComponentByClass<UArcGISLocationComponent>())
+			ArcGISCameraLocation = a->FindComponentByClass<UArcGISLocationComponent>();
```
ZackAllen, line 357:
```suggestion
		if (a->FindComponentByClass<UArcGISCameraComponent>() && a->FindComponentByClass<UArcGISLocationComponent>())
		{
			ArcGISCameraLocation = a->FindComponentByClass<UArcGISLocationComponent>();
		}
```

arash-sh: Good catch!

ZackAllen (approved):

avr757 (approved):

PR #95, opened 2024-08-22
Quality of Life Update: Sample Switcher

mmgaw113:
**Sample**

Update Sample Switcher to have a designated toggle to only show samples made in Blueprints or only show samples made in C++. This will consolidate the number of samples shown, especially as we begin porting/recreating each sample in either BP or cpp depending on the sample.

**Summary**

- Updated the sample switcher to have two new buttons for changing between showing BP only or C++ only.
- Added a new SampleDataTable for BP only Levels.
- Renamed the SampleDataTable to differentiate between the old one and the new cpp only data table.
- Update the List View to switch between BP only levels and CPP only levels based on selection.
- Made changes to the way the Level name is highlighted due to a bug found.

**Tests**

Windows

**ArcGIS Maps SDK Version**

Maps SDK public 1.6 w/ UE5.2


avr757 (approved):
UI/UX looks and works well! Nicely done.
The only thing I would add is to just make sure the sample name is the same e.g. "Third Person Controller" across both C++ and Blueprints for consistency . 

image
image

PR #99, opened 2024-09-09
Error: Fix Build Error/Build Launch Error

mmgaw113:
**Sample**

Multiple Samples Affected

**Summary**

In most of the C++ samples, we were using ConstructorHelpers to create the widgets (UI) and in some cases to load StaticMeshes or other components. This only affects packaged builds. After making a build, it would succeed, launching the build would result in an instance crash and freeze of the build. All samples functioned as expected in the editor.

**Tests**

Windows

**ArcGIS Maps SDK Version**

Maps SDK 1.6 W/UE 5.2
(I tried to test 5.4 but there are other issues that will needed to be addressed within the plugin that cause compile errors).


ZackAllen (approved):
Verified that project builds and runs correctly on Windows 10, Unreal 5.2, ArcGIS Maps SDK 1.6.

PR #104, opened 2024-09-25
Readme update

avr757:
This updates fixed the old links regarding the new API key changes.


README.md
```diff
@@ -68,7 +69,7 @@ Esri welcomes contributions from anyone and everyone. Please see our [guidelines
 
 ## Licensing
 
-Copyright 2022 Esri
+Copyright 2024 Esri
```
mmgaw113, line 72: I don't know what the rules are for updating copyright. Did Andres say we should be updating the copyright for samples?

mmgaw113 (approved):
Just address the Copyright comment if needed, otherwise looks good.

README.md
ZackAllen: Copyright is set to current year on file creation and file rename. Copyright is not changed on modification.

avr757: So the PM recommends a year range (2022-2024) for this copyright in the readme file as repo is having ongoing updates

PR #106, opened 2024-10-15
Blueprint addition: add measure blueprint sample

Jade-JadeH:
**Sample**

Note: need to use the latest daily to test this PR since the 1.6 bug was fixed recently. 

- [x] Converted measure sample to BP
- [x] Fixed a UI bug 
- [x] Simplified the logic of measure UI

Original UI structure: 

image

After simplifying

image

- [x]  Replaced the nested if-else branch such as the following as this is prone to error, with a button design: 

image

- [x] According to Adrien, implemented separate UI for each sample

image

**Checklist**

- [x] Self-review of changes
- [NA] No unnecessary includes or namespaces added
- [NA] Code follows plugin coding style
- [NA] New code and changed code has proper formatting 
- [NA] No unintentional formatting changes

**Blueprints**

**ArcGIS Maps SDK Version**


mmgaw113:
Can you please mark the buttons as images under style to get rid of the white outline around them. This needs to be done for Normal, Hovered, and Pressed effects.
Screenshot 2024-10-16 at 8 38 15 AM
Screenshot 2024-10-16 at 8 38 45 AM

mmgaw113:
If these are not being used please delete them.
Screenshot 2024-10-16 at 8 40 46 AM

mmgaw113:
Can we clean this up a bit more? It's a bit difficult to read with the connecting lines going all over the place. 

Screenshot 2024-10-16 at 8 41 19 AM

sample_project/Content/SampleViewer/Samples/Measure/Blueprints/UI_Measure.uasset
mmgaw113, line 1: Same comment for this, please straighten and clean up the file so it is more readable. Connecting lines between blueprint nodes shoudl be straight and no unnecessary nodes not in use should be present.

mmgaw113, line 1: You are getting actor of class on construct and every time one of the buttons is pressed. Please use the stored variable Measure BP, that you are creating on Event Construct.
Screenshot 2024-10-16 at 8 52 09 AM

mmgaw113:
The Info button needs to both remove the Info Card from the screen and bring it on screen. Please revert it back to its original behavior. 
ScreenRecording2024-10-16at8 48 02AM-ezgif com-video-to-gif-converter

mmgaw113:
This sample should not break the c++ Measure sample. The UI still needs to work with c++ and can not only be set to work with the Blueprint sample.
Screenshot 2024-10-16 at 8 55 45 AM

mmgaw113:
Alphabatize. 
Screenshot 2024-10-16 at 8 57 41 AM

Jade-JadeH:
> This sample should not break the c++ Measure sample. The UI still needs to work with c++ and can not only be set to work with the Blueprint sample. 

The C++ sample is currently broken on main after recent changes, as I remember you can't add any stops. I was going to submit a separate PR to fix it and also make the new UI works with it.

mmgaw113:
Please use the enhanced input system. We have deprecated using the old input system in UE. Check the hit test blueprint sample for how it is implemented.

Screenshot 2024-10-16 at 9 09 43 AM

Jade-JadeH:
> The Info button needs to both remove the Info Card from the screen and bring it on screen. Please revert it back to its original behavior. 

Nice catch! I use a flip-flop node to bring it back and forth now.

Jade-JadeH:
> Alphabetize. 
I see real-time weather and third person controller are flipped so wasn't sure whether this needs to be alphabetized or not, but now I alphabetize it

Jade-JadeH:
Ready to review again @mmgaw113 . Notice that you might need to do a clean build to check again.

mmgaw113:
Please clean this up with reroute nodes so it maintains readability.
{378FC900-151E-4414-A7F2-08FEB1781AE1}

sample_project/Source/sample_project/Measure/Measure.cpp
mmgaw113, line 1: This isn't really what I had in mind for making the UI work with both BP and CPP samples. None of our other samples are using this method for UI when the sample is c++ so I'm not sure we should do it for this one. There is a way to reuse the existing widget BP for both blueprint and c++ samples.

mmgaw113: You can create a function that replicates the functionality below and replace everywhere you call geodetic distance with this function. You can do the same thing for the `Unit` variable and for any functions in use such as `Clear line`.
{82BD2167-122F-49E5-933B-700782388EEE}

Jade-JadeH: > This isn't really what I had in mind for making the UI work with both BP and CPP samples. None of our other samples are using this method for UI when the sample is c++ so I'm not sure we should do it for this one. There is a way to reuse the existing widget BP for both blueprint and c++ samples.

For C++ sample, I was comparing between whether to make UI work using C++ or using blueprint and I opted in for using C++. You can def do it in bp, but since it's c++ I think we should do things in c++ as much as possible. In fact, I think other samples should do the same thing instead of mixing BP and C++, i.e., apart from widget design, everything is in c++.

mmgaw113: I don't necessarily agree with this. Not many people who use our samples are doing anything in c++ since they want blueprint-specific versions. Keeping the UI consistent by only using one Widget to work with 2 samples will help maintain useability and ease of use for updating samples in the future. Doing all of the widget functionality for C++ specific samples adds unnecessary complexity to the most simplistic part of our samples. It also leaves the samples code vulnerable to deprecations within our c++ code. We already need to address deprecation warning in 5.4 and above for the Feature Layer, Weather, and new Hit Test samples. I think we should keep the UI how it is currently implemented across all samples until the team can come to a conclusion on how to do it when we go through and do the Samples Re-work project.

Jade-JadeH: There are past customers using c++ as I've talked with them. 

It's a design decision about whether we should do everything in c++ for c++ sample or some in BP. I'm not saying we should go and change all other samples to do this unless we decide so on the sample meeting. 

For this particular sample, I think we should keep this scheme to show people how to do this in C++, even as a different workflow. I also don't think it's resource-worthy to change things around.

Jade-JadeH:
> Please clean this up with reroute nodes so it maintains readability. {378FC900-151E-4414-A7F2-08FEB1781AE1}

I don't like the bp lines either, I just tried again and how about this: 

image

mmgaw113:
> I don't like the bp lines either, I just tried again and how about this:
Better, but can you please move the reroute nodes below the other nodes. You can move the is valid and instructions up, and then put all re-route nodes at the bottom only coming off one reroute node.

Jade-JadeH:
@mmgaw113 ready for re-review. I didn't format some bps in case there's feedback. Once this is good I'll do formatting.

Jade-JadeH:
Use bp to do UI interaction. Also fixed an issue that @ZackAllen brought up in sample switcher when switching between bp and c++ which we don't unload and load stream level.

sample_project/Content/SampleViewer/User-Interface/wbp_SampleSelector.uasset
mmgaw113, line 1: Two things about the bug you listed that zack pointed out:
1.) it needs to be in a separate PR. Not this one. It does not directly affect this PR and should be separate
2.) It's not really a bug. I think I know what you are saying but the intention of the buttons at the top for toggling the list of BP samples and the List of CPP samples, however we do not want to be unloading the level and reloading the same level when switching. This doesn't solve the problem of the Sample being highlighted if it has the same name. Additionally, I tested it and the level never finished loading. It would never apply the API key to the map creator. 

TLDR: please revert changes to this file and we can discuss a bit more in depth how to address this. We can delay fixing it until the Samples refactor. It's not affecting anything major.

sample_project/Source/sample_project/Measure/Measure.h
```diff
@@ -18,6 +18,7 @@
 #include "Blueprint/UserWidget.h"
 #include "Components/ComboBoxString.h"
 #include "Components/SplineMeshComponent.h"
+#include "Components/TextBlock.h"
```
mmgaw113, line 21: Unnecessary Include

sample_project/Source/sample_project/Measure/Measure.cpp
```diff
@@ -247,35 +251,52 @@ void AMeasure::ClearLine()
...
+		FName VariableName(TEXT("Distance")); 
+		FProperty* text = UIWidget->GetClass()->FindPropertyByName(VariableName);
+		if (text)
+		{
+			FTextProperty* Distance = CastField<FTextProperty>(text);
+			if (Distance)
+			{
+				Distance->SetPropertyValue_InContainer(UIWidget, NewText);
+			}
+		}
```
mmgaw113, line 269: Can we invert the if statement to make it more readable? 
`if (!UIWidget)`
`{`
`     return;`
`}`

`code block originally in the if statement`

mmgaw113 (commented):
Looks good. thank you for making the changes. Just a few more smaller things. Please clean up the blueprints (remove unused nodes, add reroute points to improve readability, etc), revert the changes made to wbp_SampleSelector since it needs to be done in a separate PR and we need to look more into it, make the few small code changes and it will be ready to go.

sample_project/Content/SampleViewer/User-Interface/wbp_SampleSelector.uasset
Jade-JadeH: @ZackAllen Zack mentioned this to me the other day, so I'm assuming it needs to be fixed in this PR

The issue is regarding to switching between c++ and BP samples (so both measure, but measure bp and measure c++), the measure line doesn't get destroyed

Jade-JadeH:
> Looks good. thank you for making the changes. Just a few more smaller things. Please clean up the blueprints (remove unused nodes, add reroute points to improve readability, etc), revert the changes made to wbp_SampleSelector since it needs to be done in a separate PR and we need to look more into it, make the few small code changes and it will be ready to go.

Thanks for reviewing : D

sample_project/Content/SampleViewer/User-Interface/wbp_SampleSelector.uasset
mmgaw113: Is this for changing from the CPP to BP measure sample? Or changing the List of samples loaded? Based on the code you added to this file it it unloading and reloading the level when changing the list of samples which we should not be doing. We should only change the level that is loaded when clicking on the individual level. The CPP and Blueprint Buttons are only for changing the list of samples loaded.

If the bug you are referring to is that clicking on the Blueprints button does not get rid of the measure path, that isn't a bug. That is by design. If it isn't that, then I still don't understand what the bug is and would still like it to be a separate PR.

[This video]( https://github.com/user-attachments/assets/d0301f6a-8fd3-4336-82a6-d5840b87c963) is how it is supposed to be.

[This video](https://github.com/user-attachments/assets/662d5850-f20d-4245-b1f7-d29fd200cf90) is how it is on your branch.

As I said, please revert this change to this file. This needs to be investigated more.

mmgaw113:
Please also address the conflicts with SampleViewer.umap

sample_project/Content/SampleViewer/User-Interface/wbp_SampleSelector.uasset
Jade-JadeH: Yes, the expected behavior above is what Zack wants me to change (change to the second video). Since the two of you have different opinions on expected behavior and Zack's on vacation, maybe better to hold off until Monday?

Jade-JadeH: Do you mind hopping on a call to make sure we're on the same page?

Jade-JadeH:
> Please also address the conflicts with SampleViewer.umap

I usually do this after approval and before merging. I can do this earlier next time

mmgaw113:
No worries, I would prefer everything be update to date with main and Release branch while I am reviewing to make sure nothing is getting broken. It's a good practice to make sure nothing gets merged broken and then requiring a fix later.

Jade-JadeH:
Forgot this is going to release branch instead of main so I did some back and forth commits there, but now should be good for re-review

sample_project/Source/sample_project/Measure/Measure.cpp
mmgaw113: This one doesn't look like it got solved in the diff. I see it in your commit history but it's not here. However you don't need the `else`. You can just do 

    void AMeasure::UpdateDistance(float Value)
    {
	    if (!UIWidget)
	    {
		return;
	    }

	FText NewText = FText::AsNumber(Value);
	FName VariableName(TEXT("Distance")); 
	FProperty* text = UIWidget->GetClass()->FindPropertyByName(VariableName);
	
	    if (text)
	    {
		    FTextProperty* Distance = CastField<FTextProperty>(text);
		    if (Distance)
		    {
			Distance->SetPropertyValue_InContainer(UIWidget, NewText);
		    }
	    }
    }

mmgaw113:
Cpp Hit Test does not work anymore after address conflicts. Please add it back to the list of sub levels
{C6623247-2655-49C3-8427-2A4DAF5F848A}

mmgaw113 (approved):
Nice job.

ZackAllen (approved):

PR #119, opened 2024-11-08
bug: fix address lookup in geocoding sample

arash-sh:
**Sample**

Fix the address lookup resulting in a black screen
Use draw status to set the camera altitude when the map is ready

**Checklist**

- [x] Self-review of changes

**ArcGIS Maps SDK Version**
1.7 daily


Jade-JadeH (approved):
Confirmed that this fixes the issue both in the sample and also in the sample viewer.

Jade-JadeH:
While testing I do notice a UX improvement which is that when users enter incomplete address there's no feedback. Still trying to figure out when this happens because when I enter "1919 McIntyre Street" it doesn't take me anywhere, which I guess is because of incomplete address? Because "1919 McIntyre Street, Ann Arbor" does return a place.

but when I enter "380 new york street" which is also incomplete, it does take me to somewhere 

image

Maybe this is just because of the issue with REST API?

It would still be useful if on the instruction banner below we can show a message when this happens. 

image

avr757:
This is a minor UI thing, but will it be possible to change the result text and background to match with the UI theme (darker background, white texts w/ the special fonts)

image

ZackAllen (changes requested):
You branched from main, and this PR is bringing in other unrelated changes from `main`. Please rebase from the release branch before merging.

mmgaw113:
> This is a minor UI thing, but will it be possible to change the result text and background to match with the UI theme (darker background, white texts w/ the special fonts)
> 
> image

Andy, would you be okay with waiting to make this change until we start fixing up samples? This is how it is and has been for quite a while and changing it will be a bit complicated. Doable, but complicated. And since this sample currently doesn't work I think priority should be fixing the sample asap.

PR #124, opened 2024-11-20
Bug: Fix geocoding sample

mmgaw113:
**Sample**

Geocoding Sample

**Checklist**

- [x] Self-review of changes

**Blueprints**
N/A

**ArcGIS Maps SDK Version**

Maps SDK 1.7 w/ UE 5.3


Jade-JadeH:
Copying my comments from another PR to here: 

While testing I do notice a UX improvement which is that when users enter incomplete address there's no feedback. Although I think this can be a different PR. 

Still trying to figure out when this happens because when I enter "1919 McIntyre Street" it doesn't take me anywhere, which I guess is because of incomplete address? Because "1919 McIntyre Street, Ann Arbor" does return a place.

but when I enter "380 new york street" which is also incomplete, it does take me to somewhere 

image

Maybe this is just because of the issue with REST API?

It would still be useful if on the instruction banner below we can show a message when this happens. 

image

arash-sh:
> is just because of the issue with REST API?
> 
> It would still be useful if on the instruction banner below we can show a message when this happens.
Yes it is because that address doesn't return any candidates. You can check it by opening https://geocode-api.arcgis.com/arcgis/rest/services/World/GeocodeServer/findAddressCandidates?f=pjson&singleLine=1919%20McIntyre%20Street&token=<API-KEY>

I think it is useful to have a note (similar to the one for multiple candidates). I will leave it to @mmgaw113 and @ZackAllen whether that should be done in this PR.

ZackAllen (approved):
I consider it expected behavior. I would merge this fix as is and we can address the behavior for no address found later.

Jade-JadeH:
> I think it is useful to have a note (similar to the one for multiple candidates). I will leave it to @mmgaw113 and @ZackAllen whether that should be done in this PR.

I created this issue https://devtopia.esri.com/runtime/normandy/issues/2997 in case we forget about these things

PR #129, opened 2024-12-19
Quality of Life: Feature layer Refactor and updates

mmgaw113:
**Sample**

- Refactor the feature layer sample to clean up and reformat the code
- Remake the feature item in C++ for more consistent performance across the sample
- Write the Blueprint functions in bp_FeatureLayer in c++ in order to remove the Blueprint class (Note this deletion is why weather_wbp and BlueprintWeather_wbp have changed)
- Update UI to show properties of currently selected feature
- Add input to Feature Layer Query so users can select a feature
- Add visual Feedback via Highlight material to show the currently selected feature

**Checklist**

- [x] Self-review of changes

**Blueprints**
Feature Layer WBP: https://blueprintue.com/blueprint/939quqzf/

**ArcGIS Maps SDK Version**

Maps SDK 1.7 w/UE 5.3


Jade-JadeH:
Good to see that bp was removed from this sample, this is one of the things I want to improve on our sample repo. I see surface placement is also used. Great. 

1. I didn't checkout out the unity PR, but thought we're going to make the pins not overlapping?

image

2. It the feature 0-indexed or 1-indexed? because when I select feature 0-6 I'm expecting to see 7 features but seeing 6, I tried a couple of other numbers and see the same. 
image

3. How do I see the properties of the currently selected feature? Is it in the UI or in the detail panel? 
image

mmgaw113:
> 1. I didn't checkout out the unity PR, but thought we're going to make the pins not overlapping?

In order to make them not over lap whatsoever, the pin mesh scale would be so incredibly small it would be difficult to see without the camera being super close. With this specific data set (trees), they are clustered together which makes it hard to not have any overlapping. This is pretty similar to how the unity one looked.

mmgaw113:
> 2\. It the feature 0-indexed or 1-indexed? because when I select feature 0-6 I'm expecting to see 7 features but seeing 6, I tried a couple of other numbers and see the same.

It might be triggering the wrong portion of the If Statement that runs a forloop for Last Value - 1, I will have to look into it. But you are right, you should see 7.

mmgaw113:
> 3\. How do I see the properties of the currently selected feature? Is it in the UI or in the detail panel?

Shift + Click on a pin. The properties panel will only show up after you click on a pin and it will auto-hide the outfield drop down. Additionally, if you want to see the outfield drop-down while the properties panel is active, it will hide the properties panel.

Jade-JadeH:
> > 3. How do I see the properties of the currently selected feature? Is it in the UI or in the detail panel?
> 
> Shift + Click on a pin. The properties panel will only show up after you click on a pin and it will auto-hide the outfield drop down. Additionally, if you want to see the outfield drop-down while the properties panel is active, it will hide the properties panel.

Aw ok, I see. I can see that based on what outfield you select, it shows different kind of properties which works very well. Based on this I noticed: 

1. the dropdown menu appears to behave strangely 
UnrealEditor_svwKK21Vm3

Other comments I have are: 
1. A typo on the instruction: data set ---> dataset; I also think we should remove "we apologize for the inconvenience". 
2. Can we zoom to a point? For example, click on one point on the outliner and zoom to that point.

mmgaw113:
Unfortunately that is expected behavior :/ The issue I was running into is that if we want it to show up when the outfields drop down is up, it needs to be bound to opposite of the book value for that toggle.

I could probably check if "selected feature" != nullptr before toggling it.

mmgaw113:
I don't think we can zoom to a point based on the inspector. That functionality only works when not in play mode and it's the same for both engines. It would be really complicated to write and we'd need to do it in the viewport which would take up a lot of space. I think it would be a nice to have but not super important for the sample. Maybe a toolkit idea?

Jade-JadeH:
> I don't think we can zoom to a point based on the inspector. That functionality only works when not in play mode and it's the same for both engines. It would be really complicated to write and we'd need to do it in the viewport which would take up a lot of space. I think it would be a nice to have but not super important for the sample. Maybe a toolkit idea?

I can agree on that, I encountered this issue when I'm doing samples too.

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
```diff
@@ -203,68 +351,279 @@ void AFeatureLayer::OnResponseRecieved(FHttpRequestPtr Request, FHttpResponsePtr
...
+			}
+
+			const auto item = Cast<AFeatureItem>(featureItem);
+
+			if (!item)
+			{
+				return;
+			}
+
+			item->Index = StartValue;
```
ZackAllen, line 422: A lot of repeat code here with the block starting in 453. Any reason the for loop on line 437 doesn't cover this case?

ZackAllen (changes requested):

ZackAllen:
Behavior of the sample worked correctly for me. Pins could use more adjustment, there is more overlapping here than the Unity implementation.

avr757:
The UI look great overall, nice work!!  just a few suggestions:

1. The spacing between these menus seems inconsistent, we can improve this by moving the property info menu down a bit so that the spacing will be equal.
image

2. It seems in the "Point Layer" text the "L" is capitalized? I think we could also add a sentence to the instruction about clicking the pins using the "Shift + Click" so users will know about this new instruction.
image

3. I was a bit confused with this UX: I thought after we deselect/reselect some of the properties in the dropdown, the number of these properties will automatically updates. It turns out I have to click a different pin to see the updates? Is this intentional just out of curiosity? I just thought it would be an easier UX for the users if it automatically updates.
https://github.com/user-attachments/assets/3fdcc40b-abb0-4746-a72b-3fbb00d45a06

sample_project/Source/sample_project/Feature Layer/FeatureLayer.cpp
mmgaw113: Yes, there are 3 blocks because there are 3 possibilities for getting the features. The first of which being "Get all features". 
Second being get features from x to y (1-100 for instance) the third is specifically a fail safe. If the feature layer only has 100 feature but a user wants to get feature item 151, it will literally crash the engine with an array index out of bounds error. I am sure there is a way to simplify this code so it does not repeat as often however this is how it currently works. I will look into creating a function that takes in a min and max integer values in order and then run the 4 loops.

mmgaw113: There is also a case where start value = last value so it doesn't run a loop at all. Just gets the first value.

mmgaw113:
> 3\. I was a bit confused with this UX: I thought after we deselect/reselect some of the properties in the dropdown, the number of these properties will automatically updates. It turns out I have to click a different pin to see the updates? Is this intentional just out of curiosity? I just thought it would be an easier UX for the users if it automatically updates.
>     Video: https://github.com/user-attachments/assets/3fdcc40b-abb0-4746-a72b-3fbb00d45a06

This is expected. The behavior is identical in Unity. There isn't an easy way to make it update after changing selected properties without running it every frame on a Tick/Update function.

ZackAllen (approved):

PR #141, opened 2025-04-16
New Sample: Time of Day Sample

mmgaw113:
**Sample**

Time of Day Sample

**Checklist**

- [x] Self-review of changes

**Blueprints**

All BP are new BP

**ArcGIS Maps SDK Version**

Release 2.0 w/ UE 5.3


avr757 (changes requested):
A great start! Nice work on the UI so far! I tested the scene, the visuals look great and the UX is working pretty well :)

Here are my comments on the UI to align better with the design:

1. I noticed the both the gear icon and the button background color has a different colors than the info icon color w/ its background color, can you make the color the consistent among them? Also the gear icon appears to be larger than the info icon and it should be decreased to the same size as the info icon. After that, move the gear icon a little to the left, I think the spacing is a little too close
image
2. It appears that the moon icon has jagged edges than the sun icon? And the color seemed a little different. I would make them identical in color and size. Let me know if you need a different icon size
image
3. Just a thought - When the simulation starts, the slider is disabled which makes sense so users can't use it during that process.  I think only disabling the slider is not as obvious visually and one way to improve is to also dim the sun and moon icons with texts so it will be easier to see that the entire top part is disabled.
image
image

4. Can you increase the text size for the time to align with the one on the design? I think you could decrease the title size (Time of Day Feature Layer) just a little bit and increase the time text size.
image
image
5. For the toggle icon, I think it will be good to have a boarder if it's selected. it looks a little bit weird to not have an outline (doesn't look like a toggle box without boarder/outline)
image
image
6. Need a little bit of more margin spaces on the speed readings right margin, maybe only keep the speed value 2 decimals? 
image
7. The minus icon seems to be sharper and have jagged edges, can you make them align with the look of the plus icons? Also the fonts for the times are different, please align with the rest of the font types.
image
8. I would make the start button and the text a little larger like the design, be cautious with the margins around it after increasing its size.
image
image
9. One small UX issue I found was that, if you have the custom time selected and minimize the window, they stayed on the screen
UnrealEditor_9VACAXkBG7

sample_project/Source/sample_project/Feature Layer/LampPostQueryLayer.cpp
```diff
@@ -0,0 +1,552 @@
...
+			{
+				auto properties = Features[0]->AsObject()->GetObjectField(TEXT("properties"));
+				auto propertyFields = properties->Values;
+
+				for (auto key : propertyFields)
+				{
+					WebLink.OutFields.Add(key.Key);
+				}
+				//parse through the features in order to get individual properties associated with the features
+				for (int i = 0; i != Features.Num(); i++)
```
ZackAllen, line 223: Lines 223-289 are a complete repeat of code above. This should be refactored to not have such a large section of repeat code.

sample_project/Source/sample_project/Feature Layer/LampPostQueryLayer.cpp
```diff
@@ -0,0 +1,552 @@
...
+	{
+		return;
+	}
+
+	if (bGetAllFeatures)
+	{
+		auto index = 0;
+		for (auto featureData : FeatureData)
+		{
+			auto featureItem = GetWorld()->SpawnActor(ALampPostItem::StaticClass());
```
ZackAllen, line 322: 322-348, 355-385, and 520-550 all have the same code. Refactoring is needed to address this.

sample_project/Source/sample_project/Feature Layer/LampPostQueryLayer.h
```diff
@@ -0,0 +1,114 @@
...
+	UPROPERTY(BlueprintReadWrite, EditAnywhere)
+	FLink WebLink;
+
+private:
+	static void AddAdditionalMaterial(const ALampPostItem* Item, UMaterialInstance* Material);
+	void OnResponseReceived(FHttpRequestPtr Request, FHttpResponsePtr Response, bool bConnectedSuccessfully);
+	void GetMapComponent();
+	static void RemoveAdditionalMaterial(const ALampPostItem* Item);
+
+	//UPROPERTY(BlueprintReadWrite, EditAnywhere, meta = (AllowPrivateAccess))
```
ZackAllen, line 95: Commented property needs removed.

avr757:
Almost there!
1. smaller case “O” in this case to be consistent with other title formats
image
2. Not sure why the moon icon still looks larger visually, let make the size to x=25, y=28 and try it out?
image
3. Let's make the slider centered, and decrease the speed text size to the speed value text size. Because the slider is not centered, that's why the speed value text is more on the right. Are you able to make it to 2 decimals? I feel like 3 decimals aren't too helpful and it takes space.
image
4. Match the these fonts (time value, and start/stop) to other ones, I think these are in "Roboto" and should be the other one.
image
5. Make the start button a little longer as the design
image
6. More margin spaces on the bottom, and line up the last sentence "of different lighting and atmosphere effects."
image

ZackAllen (approved):

avr757 (approved):
Looking great, thanks!

PR #146, opened 2025-05-19
Bug: Fix Stream Layer Crash

mmgaw113:
**Sample**

Bug with Stream layer specifically in a build that would cause a crash from too many UObjects in the Scene. I narrowed it down to this was being caused by predicting the point of every plane in the scene every tick. Doing so was creating a new ArcGISPoint every frame for every plane (100) and would eventually crash after about a minute. The solution was to make it so only one plane is being updated at a time based on which plane has been selected by the users.

**Checklist**

- [ ] PR title follows convention - `keyword: Short description of change` 
- [ ] PR targets the correct branch 
- [ ] New Blueprints follow Unreal Naming Convention 
- [ ] Self-review of changes
- [ ] There are no warnings related to changes
- [ ] No unrelated changes have been made to any other code or project files
- [ ] No unnecessary includes or namespaces added
- [ ] Code follows plugin coding style
- [ ] New code and changed code has proper formatting 
- [ ] No unintentional formatting changes
- [ ] Commits have descriptive titles

**Blueprints**
StreamLayer_wbp (added tick to the blueprint, nothing else.

**ArcGIS Maps SDK Version**

2.0 w/ UE 5.3


Jade-JadeH:
Based on the issue description, my question is that this is a functionality change to the original samples (correct me if I'm wrong), if the user wants to see flight trajectories of a group of planes, which is a very likely scenario, and I think that's the original purpose of this sample, they won't be able to do that anymore. Plane simulation is a narrow field but I can imagine users who work in transportation department for example, they might want to see all buses in LA simultaneously in order to dispatch buses. But I also need to think what's a better way to do this, we definitely don't want to update all planes every frame, but we also shouldn't just be updating one plane.

Jade-JadeH (changes requested):
see comments

arash-sh (approved):
While I agree that the behavior change is significant and undesirable. I lean toward merging this PR to avoid a fast crash and creating another issue to look into resolving the underlying issue (which may not be possible without changing the plugin)

sample_project/Source/sample_project/Stream Layer/StreamLayerQuery.cpp
```diff
@@ -121,19 +121,7 @@ void AStreamLayerQuery::BeginPlay()
...
-{
-	for (auto Plane : planeData)
-	{
-		if (Plane.Value)
-		{
-			Plane.Value->PredictPoint(DeltaSeconds * 1000);
-			Plane.Value->LocationComponent->SetPosition(Plane.Value->PredictedPoint);
-		}
-	}
-}
```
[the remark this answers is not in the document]
mmgaw113: It got moved to the Widget BP. When you click on a plane from the drop down it gets the name of it and starts updating the position. I don't think there is an easy way to keep it in C++ since the widget is entirely blueprints.

ZackAllen (approved):

sample_project/Source/sample_project/Stream Layer/PlaneController.cpp
```diff
@@ -15,8 +15,12 @@
 
 
 #include "PlaneController.h"
-#include "ArcGISMapsSDK/BlueprintNodes/GameEngine/Geometry/ArcGISSpatialReference.h"
 #include "DeadReckoning.h"
+#include "ArcGISMapsSDK/Actors/ArcGISMapActor.h"
+#include "ArcGISMapsSDK/API/GameEngine/Geometry/ArcGISSpatialReference.h"
+#include "ArcGISMapsSDK/API/GameEngine/View/ArcGISView.h"
+#include "ArcGISMapsSDK/Components/ArcGISMapComponent.h"
```
arash-sh, line 22: Nitpick - order

sample_project/Source/sample_project/Stream Layer/PlaneController.cpp
```diff
@@ -53,9 +61,11 @@ void APlaneController::PredictPoint(double intervalMilliseconds)
...
 	FeatureData.PredictedPoint.Y = drPoint[1];
 	FeatureData.PredictedPoint.Z = currentPoint[2];
-	PredictedPoint = UArcGISPoint::CreateArcGISPointWithXYZSpatialReference(
-		FeatureData.PredictedPoint.X, FeatureData.PredictedPoint.Y, FeatureData.PredictedPoint.Z,
-		UArcGISSpatialReference::CreateArcGISSpatialReference(4326));
+	PredictedPoint = Esri::GameEngine::Geometry::ArcGISPoint(FeatureData.PredictedPoint.X,
+		FeatureData.PredictedPoint.Y, FeatureData.PredictedPoint.Z,
+		Esri::GameEngine::Geometry::ArcGISSpatialReference(4326));
+	auto newLocation = MapComponent->GetView()->APIObject->GeographicToWorld(PredictedPoint);
+	predictedLocation = MapComponent->ToEnginePosition(newLocation);
```
arash-sh, line 68:
```suggestion
	predictedLocation = MapComponent->ToEnginePosition(MapComponent->GetView()->APIObject->GeographicToWorld(PredictedPoint));
```

sample_project/Source/sample_project/Stream Layer/PlaneController.h
```diff
@@ -76,19 +76,25 @@ class SAMPLE_PROJECT_API APlaneController : public AActor
 {
 	GENERATED_BODY()
 
+protected:
+	virtual void BeginPlay() override;
+	
 public:
 	APlaneController();
+	void GetMapComponent();
+	UFUNCTION(BlueprintCallable)
```
arash-sh, line 85: No need for this to stay ufunction, right?

arash-sh (approved):

Jade-JadeH (approved):

PR #152, opened 2025-07-01
Rename geometry.PNG to geometry.png

avr757:
**Sample**
When public doc autosource the readme file in GitHub repo with images, it's case sensitive and won't render .PNG, it must take .png with smaller case

image

Preview won't render image:

image


tTakeda09:
Yeap. Samples images are still case sensitive.
I see you forgot to update the image file path on the MD file...

PR #158, opened 2025-08-19
New sample: Flythrough

Jade-JadeH:
**Sample**

As discussed in the sample meeting, this sample has the following features: 

- [ ] Support predefined Camera Flythrough Path or add current position to path 

- [ ] Smooth Position & Rotation Interpolation

-Uses vector interpolation for position and `FindLookAtRotation` for auto camera path calculation .

- [ ] Toggle Between Modes

-Flythrough Mode – Automated camera movement along the predefined path.

-Navigation Mode – Switch to free camera (Pawn) control to explore the map.

- [ ] Wait till map is loaded

-Uses `ArcGISView.GetDrawStatus()` to ensure the map is fully rendered before starting the flythrough.

- [ ] Some level of obstacle of avoidance 
It calculates a mid point and go through two arcs: start->mid point, and mid point->end; When the start and end are too close to each other with drastic height difference, this mechanism sometimes fails. 

**Checklist**

- [ ] PR title follows convention - `keyword: Short description of change` 
- [ ] PR targets the correct branch 
- [ ] New Blueprints follow Unreal Naming Convention 
- [ ] Self-review of changes
- [ ] There are no warnings related to changes
- [ ] No unrelated changes have been made to any other code or project files
- [ ] No unnecessary includes or namespaces added
- [ ] Code follows plugin coding style
- [ ] New code and changed code has proper formatting 
- [ ] No unintentional formatting changes
- [ ] Commits have descriptive titles

**Blueprints**

**ArcGIS Maps SDK Version**


mmgaw113:
Does the sample auto save points you created? Can we make that an setting/option? I don't think everyone will want that. Additionally, could we make it a key/input to save rather than auto save by default?

Can we also add a way to easily clear the waypoints added? From what I could tell there isn't currently a way to clear the waypoints you created.

Jade-JadeH:
@mmgaw113 


> Does the sample auto save points you created? Can we make that an setting/option? I don't think everyone will want that. Additionally, could we make it a key/input to save rather than auto save by default?
> 
> Can we also add a way to easily clear the waypoints added? From what I could tell there isn't currently a way to clear the waypoints you created.

I added these functionalities, you can re-review now.

Jade-JadeH:
@avr757 I addressed your UX comments, @mmgaw113 readme is added so ready for re-review

mmgaw113:
Flew right into the building and just stopped 😞 
https://github.com/user-attachments/assets/9ccba26d-876b-4f20-a93d-8c6ffe724ee0

Jade-JadeH:
> Flew right into the building and just stopped 😞 https://github.com/user-attachments/assets/9ccba26d-876b-4f20-a93d-8c6ffe724ee0

The obstacle avoidance fails at time: I added it in the readme. 
When traveling between two stop points, the system does a raytracing to determine the obstacle on the flythrough path, then computes a midpoint above the obstacle, and traverses two arcs (start → mid, mid → end) to reduce chances of intersecting terrain/objects. Note: When start and end are very close but have a large height difference, this heuristic can fail.

One way to address is to allow pawn movement during flythrough mode, so that you can unstuck yourself. I purposely disabled it according to @avr757 feedback though....

avr757 (approved):
Thanks @Jade-JadeH for your effort in improving the UI/UX after a few rounds of back-and-forth. Tested it out and UI, UX, visuals all look great now! Thanks again, and great job! :D

sample_project/Content/SampleViewer/Samples/Flythrough/readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+3. Wait for the map to fully render. The app starts with two stop points. 
+
+4. You can toggle between free navigation mode and flythrough mode by pressing `Tab`. In the free navigation mode, you can press `T` to add the current camera position as a stop point. (Tip: Click in the viewport to ensure it has input focus before pressing.) You can also open the blueprint `Flythrough_BP` to add stop points. After adding the stop points, you can switch back to the flythrough mode to fly through these points. Clear the stop points by clicking the clear button. 
+
+5. Change the flythrough speed by changing the variable `speed`. 
+
+6. You can toggle autosave on and off. When autosave is on, the app saves the stop points you added to the disk so you can reload them next time. 
+
+## How it works
```
avr757, line 21: One quick thought, in the "how it works", should we mention where the saved file is?

